PR #1, opened 2017-05-27
Fix for evaluation of distribution release and default provider.

tabowling:
@mlichvar  @larskarlitski   I must admit I did not test this, but the prior logic did not seem to properly account for RHEL version >= 6.8.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -17,7 +17,7 @@
 
 - name: Select NTP implementation
   set_fact:
-    ntp_implementation: "{{ 'ntp' if ansible_distribution in ['RedHat', 'CentOS'] and ansible_distribution_version|version_compare('6.8', '<') else 'chrony' }}"
+    ntp_implementation: "{{ 'ntp' if ansible_distribution in ['RedHat', 'CentOS'] and ansible_distribution_major_version|version_compare('7', '<') else 'chrony' }}"
   when: sync_mode != 2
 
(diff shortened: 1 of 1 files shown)


mlichvar:
What exactly did not work? The code was supposed to select chrony on 6.8 and later.
If I understand it correctly, the new code would select ntp instead.

larskarlitski:
This fails on rhel-6 with `No package matching 'ntp' found available, installed or updated` and on rhel-7 with `No package matching 'linuxptp' found available, installed or updated`.

larskarlitski:
> So, this is not a bug in the role, but in the system images used for testing.

I'm sorry, I don't understand. Shouldn't these roles work without errors on the released images?

mlichvar:
The Fedora and CentOS images work.

If the RHEL image doesn't come with the linuxptp package and it's not possible to install packages from a repository, what should the role do if it is configured to synchronize to a PTP domain?

larskarlitski:
The point of these roles is to make common tasks possible on different versions of operating systems. Surely there was a (different?) way to synchronize to a PTP domain on RHEL 6?

Why is not possible to install packages?

mlichvar:
There are no configured repositories in /etc/yum.repos.d, so the ansible package module (which calls yum install) fails.

larskarlitski:
Sorry, I misread that above. I guess we need to register these images to make repositories available.

PR #3, opened 2018-05-16
Move tests to /tests directory and rename to new naming scheme

tyll:
Ansible Galaxy and
https://fedoraproject.org/wiki/CI/Standard_Test_Interface suggest to use
a /tests directory and names matching test*.yml


mlichvar:
Looks good to me.

tyll:
[citest bad]

tyll:
The CI works with this change, any failures in the role itself should be fixed in a different PR.

PR #6, opened 2018-05-18
Improve tests to check if configured sources are actually used

mlichvar:


mlichvar:
Some of the new tests work only on machines with SW/HW timestamping and currently don't run in the CI. It would be nice if we could tell the standard-inventory-qcow2 script to configure the VM with the e1000 driver, which supports SW timestamping.

tyll:
Best is to file a new issue in the upstream project for the inventory file: https://pagure.io/standard-test-roles/issues

mlichvar:
Ok. I've filed https://pagure.io/standard-test-roles/issue/196

pcahyna:
Not sure why the centos-6 test failed, but it does not seem to be related to the changes. The error message is from yum and it reads "Failure talking to yum: Cannot retrieve repository metadata (repomd.xml) for repository: base. Please verify its path and try again". I tested it locally and the tests pass. I am merging the PR.

pcahyna:
It seems that the support for configuring the NIC driver is now in standard-test-roles. What is needed to use it in our tests?

mlichvar:
I think we just need to enable the e1000 driver. Where is the CI integration for this repository located?

tyll:
It is the test harness: https://github.com/linux-system-roles/test-harness

mlichvar:
Per https://pagure.io/standard-test-roles/issue/196#comment-516970 we just need to add a file to the container that would set the NIC model. I'm not sure where it should be placed. I wasn't able to get current version of standard-inventory-qcow2 working locally to test it.

PR #9, opened 2018-07-19
Use ntp on all versions of RHEL6 (rhbz#1597684)

mlichvar:


pcahyna:
lgtm, merging. Thanks!

PR #10, opened 2018-07-20
Use facts instead of phc_ctl to check for HW timestamping

mlichvar:
Recent versions of ansible include support for SW/HW timestamping in facts, which can be used instead of the phc_ctl command.

https://github.com/tedelhourani/ansible/commit/d2f9359ecc1e85fa3d9911cc2856812d7c38037e

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -32,16 +32,9 @@
   package: name=linuxptp state=present
   when: sync_mode != 1
 
-- name: Run phc_ctl on PTP interface
-  command: phc_ctl -q {{ ptp_domains[0].interfaces[0] }}
-  register: phc_ctl_output
-  changed_when: False
-  when: sync_mode == 2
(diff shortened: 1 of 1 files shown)


pcahyna:
I see this is since 2.4.

mlichvar:
It looks like the test is failing on systems that don't have ethtool installed.

pcahyna:
It needs then to install ethtool (I think the network role does it already) and refresh the facts (prior art for this is https://github.com/linux-system-roles/selinux/pull/20/commits/9ac7cf47990c845ae61dfc470da4a729ee30b883 )

mlichvar:
That doesn't sound like an improvement to me. The idea was to remove the task running phc_ctl. If it needs to be replaced with one installing ethtool and another refreshing the facts, I guess we should rather drop this change.

pcahyna:
I think you are right.

PR #11, opened 2018-07-20
Make the provider configurable.

pcahyna:
Users now can set the ntp_provider variable. If they don't, the default remains
as before.
https://bugzilla.redhat.com/show_bug.cgi?id=1597693

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -15,10 +15,10 @@
     sync_mode: 3
   when: sync_mode is not defined
 
-- name: Select NTP implementation
+- name: Select NTP provider
   set_fact:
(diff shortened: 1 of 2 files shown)


mlichvar:
The change looks good to me, but I think we will need a different approach if we want to also detect the currently installed/running NTP daemon. I'm experimenting with a module.

pcahyna:
The idea is that autodetection would just extend this approach, rather than being radically different. I.e. it would set the variable `ntp_provider_current` and a fail task would be added which would compare `ntp_provider_current` to `ntp_provider_default` if `ntp_provider` is not given by the user.
By the way, the variable should be called `timesync_ntp_provider` (namespacing).

mlichvar:
FWIW, the PR #12 moves the system-specific selection to the module.

PR #12, opened 2018-07-23
Select currently active/enabled NTP provider by default (rhbz#1597692)

mlichvar:
This uses a simple module written in shell for selecting the provider by currently active/enabled service, or falling back to the system-specific provider. The user can override the selection by setting the ntp_provider variable.

### DIFF
--- a/library/timesync_provider.sh
+++ b/library/timesync_provider.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+# WANT_JSON
+
+is_service_enabled() {
+	local name=$1 runlevel prev_runlevel
+
+	read -r prev_runlevel runlevel < <(runlevel)
+
+	systemctl is-enabled $name.service &> /dev/null || \
+		chkconfig --list $name 2>/dev/null | grep -q "$runlevel:on"
+}
+
+is_service_active() {
+	local name=$1
+
+	systemctl is-active $name.service &> /dev/null || \
+		service $name status &>/dev/null
+}
+
+get_current_ntp_providers() {
+	is_service_active chronyd || is_service_enabled chronyd && echo chrony
+	is_service_active ntpd || is_service_enabled ntpd && echo ntp
+}
+
+get_default_ntp_provider() {
+	[ -e /etc/os-release ] && . /etc/os-release &> /dev/null
+
+	case "$ID" in
+		rhel|centos)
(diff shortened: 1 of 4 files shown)


library/timesync_provider.sh
```diff
@@ -0,0 +1,57 @@
...
+	systemctl is-active $name.service &> /dev/null || \
+		service $name status &>/dev/null
+}
+
+get_current_ntp_providers() {
+	is_service_active chronyd || is_service_enabled chronyd && echo chrony
+	is_service_active ntpd || is_service_enabled ntpd && echo ntp
+}
+
+get_default_ntp_provider() {
```
pcahyna: I would omit this function and do the default provider detection in the task file as before, for several reasons.

1. Consistency - there are several other places in linux-system-roles that check `ansible_distribution` or `ansible_distribution_version`, but nowhere else one parses `/etc/os-release`.
2. The module should do one thing and do it well. In this form the module does several thing: 1. checks for the OS default 2. checks for the current situation 3. implements the algorithm of choosing between them. I would prefer the module to perform only the second - checking the current provider - as this seems to be difficult to do without a module (for the record, as you found out, the `service_facts` module is currently insufficient for the task - it does not return enabled/disabled status of services and also is still _preview_ which is not ideal, so one needs to implement a module for the task). The other parts are easy to do in the task file so it would be better to do them in the task file.
3. Separating the OS default detection from the current service detection would enable the user to do
```yaml
ntp_provider: "{{ ntp_provider_default }}"
```
to always set the OS default regardless of the previous state. This is a reasonable thing to want and it is not possible in the current situation.

pcahyna: I implemented the suggested changes in #13

pcahyna:
Superseded by #13 which contains it with some changes.

pcahyna:
@mlichvar I just saw a possible solution to the problem of checking service status: using the service module in check mode:
```yaml
- service: name=myservice enabled=yes
  check_mode: true
  register: isit

- fail: msg=its not enabled
  when: isit is changed 
```
https://groups.google.com/d/msg/ansible-project/RmsaUial1L8/2vdkuR1NBgAJ

mlichvar:
Interesting, but it feels a bit hackish to me. I suspect it would cause a playbook to report non-zero changes even when nothing has actually changed.

pcahyna:
See also https://github.com/ansible/ansible/issues/43642

PR #15, opened 2018-07-24
 Add empty timesync_ntp_provider to defaults 

mlichvar:
Set the variable in defaults/main.yml to an empty string, so all input variables of the role are included there.

This PR includes #14 to avoid merge conflicts.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -4,3 +4,4 @@ timesync_ptp_domains: []
 timesync_dhcp_ntp_servers: false
 timesync_step_threshold: -1.0
 timesync_min_sources: 1
+timesync_ntp_provider: ''
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -17,12 +17,12 @@
 
 - name: Determine current NTP provider
   timesync_provider:
-  when: timesync_mode != 2 and (timesync_ntp_provider is not defined or timesync_ntp_provider == '')
+  when: timesync_mode != 2 and timesync_ntp_provider == ''
 
 - name: Select NTP provider
   set_fact:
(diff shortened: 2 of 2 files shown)


pcahyna:
Is it really necessary for anything? I would prefer to leave the variable undefined if not set by the user.

mlichvar:
It's not really necessary, but I think it's good for consistency. All input variables are defined and they are all listed in the same file.

tyll:
[citest bad]

tyll:
[citest all]

tyll:
[citest all]

tyll:
[citest bad]

tyll:
[citest bad]

tyll:
One might expect the defaults file to specify the API of a role, since it contains default values for the variables that are expected to be overridden by a role. Therefore it might be a good idea to specify the variable there.

tyll:
This is now needed for Satellite - @mlichvar there is now a conflict, can you please resolve it? Maybe wait a little since I might be merging other PRs as well.

pcahyna:
Indeed, this helps Foreman/Satellite, cf: https://github.com/linux-system-roles/selinux/pull/43

I would like to settle on a common convention, employed in that PR: the special value used to indicate that a setting should be preserved should be null (instead of an empty string). Would that be ok?

tyll:
> Indeed, this helps Foreman/Satellite, cf: [linux-system-roles/selinux#43](https://github.com/linux-system-roles/selinux/pull/43)
> 
> I would like to settle on a common convention, employed in that PR: the special value used to indicate that a setting should be preserved should be null (instead of an empty string). Would that be ok?

This sounds ok on first thought. Please note, that his PR is already merged, so we need a new one to do this. Also we should document this decision in a design decisions or best practices document and have the discussion there. This makes it easier to discover than a PR for just one role.

IMHO the document could be added to https://github.com/linux-system-roles/linux-system-roles.github.io

PR #17, opened 2018-07-25
Rename timesync_ntp_provider_default to timesync_ntp_provider_os_default

pcahyna:
for consistency with network, see https://github.com/linux-system-roles/network/issues/71

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -21,7 +21,7 @@
 
 - name: Select NTP provider
   set_fact:
-    timesync_ntp_provider: "{{ timesync_ntp_provider_default if timesync_ntp_provider_current == '' else timesync_ntp_provider_current }}"
+    timesync_ntp_provider: "{{ timesync_ntp_provider_os_default if timesync_ntp_provider_current == '' else timesync_ntp_provider_current }}"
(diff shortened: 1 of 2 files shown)


mlichvar:
Makes sense to me.

tyll:
Should this maybe be `timesync_provider_os_default` The network role also uses `$ROLENAME_provider_os_default`.

mlichvar:
The "ntp" makes it clear that it configures the NTP provider. There is a theoretical possibility that multiple PTP providers will be supported and a variable may be needed to allow the user to select between them.

tyll:
Are NTP and PTP configured at the same time or will only one of them be running at the same time? If it is only one of them, then `timesync_provider_os_default` can point to the provider that is used for the system that is currently configured, e.g. chrony when ntp is used or linuxptp when ptp is active.

mlichvar:
The role allows PTP and NTP to be configured at the same time. In this case the clock is controlled by chronyd/ntpd and ptp4l/phc2sys are configured as SHM reference clocks for the PTP time sources.

tyll:
[¢itest bad]

tyll:
[citest all]

PR #21, opened 2018-08-29
Add test for setting the default NTP provider for OS release

vcrhonek:


mlichvar:
This is a great idea.

It looks like the test assumes that both providers are available. That may not be true for all supported releases. Maybe the test should try to install both providers first and if that fails, skip the rest of the test?

vcrhonek:
Thanks, that makes sense, implemented. I wanted to use meta/when, but it seems that I hit:
https://github.com/ansible/ansible/issues/35890

Workarounded with set_fact.

mlichvar:
Looks good to me. Thanks.

The two commits could be squashed before merging.

tyll:
FYI: in #24 I changed the symlink/name of the role to linux-system-roles.timesync for the tests, so this would need be updated here or there as well before both PRs are merged.

tyll:
[citest]

tests/tests_ntp_provider6.yml
```diff
@@ -0,0 +1,72 @@
+
+- name: Configure NTP with OS release non-default provider and then change it to the default provider
+  hosts: all
+  vars:
+    is_dist: "{{ ansible_distribution in ['RedHat', 'CentOS'] }}"
+    is_vers: "{{ ansible_distribution_version is version_compare('7.0', '<') }}"
```
tyll: AFAICS there could be a single `is_el6` variable that would make it easier to understand:
```
is_el6: "ansible_distribution in ['RedHat', 'CentOS'] and ansible_distribution_version is version('7.0', '<')
```

Also `version_compare` is now deprecated. Another improvement would be to use `ansible_distribution_major_version` and compare it to `'6'`.

tests/tests_ntp_provider6.yml
```diff
@@ -0,0 +1,72 @@
...
+  tasks:
+  - name: Check for availability of both NTP providers
+    package: name={{ item }} state=present
+    register: package_install
+    ignore_errors: yes
+    with_items:
+      - chrony
+      - ntp
+
+  - set_fact:
```
tyll: A reference to the Ansible bug in a comment would be nice to explain why `with_items` cannot be used directlly,

tests/tests_ntp_provider6.yml
```diff
@@ -0,0 +1,72 @@
...
+    when: not both_avail
+
+  - name: Remove NTP providers
+    package: name={{ item }} state=absent
+    with_items:
+      - chrony
+      - ntp
+
+  - include_role:
+      name: timesync
```
tyll: This needs to be changed to `linux-system-roles.timesync` now.

tests/tests_ntp_provider6.yml
```diff
@@ -0,0 +1,72 @@
...
+    when:
+      - is_dist and is_vers
+
+  - name: Check ntpd service
+    shell: ntpq -c rv | grep 'associd=0'
+    when:
+      - not is_dist or not is_vers
+
+  - include_role:
+      name: timesync
```
tyll: This needs to be changed to `linux-system-roles.timesync` now.

vcrhonek:
Thanks, I've changed the role name and incorporated your other suggestions.

tests/tests_ntp_provider6.yml
```diff
@@ -0,0 +1,73 @@
+
+- name: Configure NTP with OS release non-default provider and then change it to the default provider
+  hosts: all
+  vars:
+    is_el6: "ansible_distribution in ['RedHat', 'CentOS'] and ansible_distribution_version is version('7.0', '<')"
```
mlichvar: Would it make sense to rename this variable to something less OS specific? Maybe `is_ntp_default`?

tyll: Good idea, it will make it more obvious why there is this test. @vcrhonek it would be great if you could change this.

vcrhonek: I agree, updated.

tyll:
[citest bad]

PR #22, opened 2018-10-05
Add provision file to tests

yontalcar:
Some of the tests work only on machines with HW/SW timestamping.
Add provision.fmf to configure QEMU to emulate e1000 card, which
supports timestamping.

This needs https://github.com/linux-system-roles/test-harness/pull/41 to have any effect on our CI.


mlichvar:
Thanks for your work on this.

Just a nit for the commit message. It's the kernel drivers that support or not support SW timestamping, not the HW. The virtio driver does support SW timestamping, but not in the older kernels from RHEL7. With e1000 we can run tests that require SW timestamping on more systems.

tyll:
> Thanks for your work on this.
> 
> Just a nit for the commit message. It's the kernel drivers that support or not support SW timestamping, not the HW. The virtio driver does support SW timestamping, but not in the older kernels from RHEL7. With e1000 we can run tests that require SW timestamping on more systems.

I updated the commit message to reflect this.

PR #24, opened 2018-10-10
Use linux-sytem-roles.timesync as name for role

tyll:
This makes the tests/examples match the name when the role is installed
via ansible-galaxy.


tyll:
[citest pending]
[citest bad]

tyll:
How would you like to make the role a parameter? IMHO it will be a different issue that is not specific to this role, so maybe let's discuss it somewhere else. Maybe at least in an issue here if we do not have a better place.

PR #25, opened 2018-10-23
Add example playbooks

tyll:
Updated examples from https://git.centos.org/tree/rpms!rhel-system-roles.git/c7-extras/SOURCES


pcahyna:
@mlichvar you are right, and maybe that was the reason those examples were downstream-only (I don't remember anymore...)

PR #30, opened 2019-04-10
Fix yamllint error

vcrhonek:
Fixes indentation in role's metadata:
--> Executing Yamllint on files found in timesync/...
    timesync/meta/main.yml
      9:3       error    wrong indentation: expected 4 but found 2  (indentation)

PR #28, opened 2018-12-07
Fix role to work in check mode

mlichvar:
Force the command tasks to run in normal mode in order to allow the role
to run in check mode.

Also, change "False" to "no" for consistency.

Fixes #27.

PR #33, opened 2019-04-16
Configure Molecule and Travis CI (this PR should trigger build in CI)

vcrhonek:
Configures Molecule to do basic checks (lint/syntax, run, idempotence, Ansible check mode functionality) of the role with default parameters (tests/tests_default.yml) using Travis CI and centos6/7 docker containers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+---
+language: python
+services:
+  - docker
+install:
+  - pip install molecule docker
+script:
+  - molecule --version
+  - ansible --version
+  - molecule lint
+  - molecule syntax
+  - molecule test
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 timesync
 ========
+[![Travis Build Status](https://travis-ci.org/linux-system-roles/timesync.svg?branch=master)](https://travis-ci.org/linux-system-roles/timesync)
 
 This role installs and configures an NTP and/or PTP implementation to operate
 as an NTP client and/or PTP slave in order to synchronize the system clock with
--- a/molecule/default/Dockerfile.j2
+++ b/molecule/default/Dockerfile.j2
@@ -1,0 +1,14 @@
+# Molecule managed
+
+{% if item.registry is defined %}
+FROM {{ item.registry.url }}/{{ item.image }}
+{% else %}
+FROM {{ item.image }}
+{% endif %}
+
+RUN if [ $(command -v apt-get) ]; then apt-get update && apt-get install -y python sudo bash ca-certificates && apt-get clean; \
+    elif [ $(command -v dnf) ]; then dnf makecache && dnf --assumeyes install python sudo python-devel python2-dnf bash && dnf clean all; \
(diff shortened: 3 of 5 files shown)


vcrhonek:
https://travis-ci.org/linux-system-roles/timesync

vcrhonek:
To make the build pass, we need to resolve PR#28, PR#30 and ISSUE#31.

pcahyna:
@vcrhonek it looks like Molecule stops on first error. Therefore, one does not see that one needs to merge #28 until one merges #30 and reruns the CI again. Is there a way to make Molecule show all errors, i.e. execute all actions?

vcrhonek:
@pcahyna I don't think so. I believe we've already discussed it in the context of skipping lint warnings:
https://github.com/ansible/molecule/issues/1308

We can run multiple molecule commands in .travis.yml to see errors from multiple actions.

But the test matrix differs for each command. For example, "molecule converge" runs dependency, create, prepare, converge (thus it won't fail on lint/syntax errors), "molecule check" runs  destroy, dependency, create, prepare, converge, check, destroy (thus it won't fail on lint/syntax errors, but could fail during converge and skip check because of that). I don't know whether test matrix is configurable for these commands in the same manner as it is for "molecule test" (and don't see it with quick look into Molecule's documentation).

pcahyna:
@vcrhonek I understood that lint warnings issue as being a bit different: I thought that it was about warning not being treated as errors, while here I would like Molecule to continue to the next step, even if something is treated as an error.

> But the test matrix differs for each command. For example, "molecule converge" runs dependency, create, prepare, converge (thus it won't fail on lint/syntax errors), "molecule check" runs destroy, dependency, create, prepare, converge, check, destroy (thus it won't fail on lint/syntax errors, but could fail during converge and skip check because of that). 

I think it is fine for "molecule check" to skip check if converge fails. So "molecule check" may be the right thing for us.

> I don't know whether test matrix is configurable for these commands in the same manner as it is for "molecule test" (and don't see it with quick look into Molecule's documentation).

Maybe we don't need it, wouldn't the default sequence for "molecule check" be fine for us?

vcrhonek:
@pcahyna I see. We can run "molecule lint", "molecule syntax" and "molecule check" then. The first two will cover lint/syntax separately, the third will cover running the role and running the role in check mode if the first step succeeds.

vcrhonek:
I forgot idempotence. Will think about it.

pcahyna:
@vcrhonek doesn't this mean that you can configure the sequence (matrix) for all commands, not just "test"? https://molecule.readthedocs.io/en/latest/configuration.html#scenario ? There is check_sequence, converge_sequence

vcrhonek:
> @vcrhonek doesn't this mean that you can configure the sequence (matrix) for all commands, not just "test"? https://molecule.readthedocs.io/en/latest/configuration.html#scenario ? There is check_sequence, converge_sequence

Yes, it does. I wasn't able to find.

pcahyna:
One can then configure either the converge or check command to run all of converge, check and idempotence, or configure "test" to do it (and skip syntax checking).

vcrhonek:
> One can then configure either the converge or check command to run all of converge, check and idempotence, or configure "test" to do it (and skip syntax checking).

Yes, that's exactly what I did locally (reconfigured test to do converge, idempotence and check). I'm going to update PR.

PR #39, opened 2019-05-31
Run tests_default both in normal and in check mode

i386x:
Add a playbook that works as a wrapper that runs tests of role
with default values of role variables both in normal and in
check mode.

### DIFF
--- a/tests/inventory.yaml.j2
+++ b/tests/inventory.yaml.j2
@@ -1,0 +1,1 @@
+{{ { 'all': { 'hosts': hostvars } } | to_yaml }}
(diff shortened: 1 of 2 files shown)


tests/_tests_default.yml
```diff
@@ -1,6 +1,7 @@
 
 - name: Ensure that the role runs with default parameters
   hosts: all
+  check_mode: "{{ run_in_check_mode }}"
```
pcahyna: `| d(omit)` ? not sure whether it will work

pcahyna:
you are removing tests_default so you need to add a new tests_default which would import _tests_default. Why not keep the name though?

tests/_tests_default.yml
pcahyna: no

PR #23, opened 2018-10-10
Use `is version` instead of `|version_compare`

tyll:
Using tests as filters is deprecated and `version_compare` was renamed
to `version` in Ansible 2.5.

PR #41, opened 2019-06-03
Add default vars test

i386x:
Add test that checks the role defines all its parameters in defaults.

### DIFF
--- a/tests/tests_default_vars.yml
+++ b/tests/tests_default_vars.yml
@@ -1,0 +1,20 @@
+---
+- name: Ensure that the role declares all paremeters in defaults
+  hosts: all
+
+  tasks:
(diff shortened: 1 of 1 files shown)


pcahyna:
This is in analogy with https://github.com/linux-system-roles/selinux/pull/43 and tests the problem fixed in PR #15

mlichvar:
There is a typo in `declares all paremeters`. Other than that it looks good to me. Thanks.

pcahyna:
it should use `version` instead of `version_compare`: #23 https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#version-comparison

pcahyna:
#57 fixes that#57

PR #46, opened 2019-07-19
Use timesync_ntp_peers to configure peers

hairmare:
This allows me to configure a subset of the machines as a locally peering ntp cluster that can still keep the time when we have network connectivity issues.
The machines that have a peering setup like this are the only machines on the network are allowed to connect to remote NTP servers. In my case they will have pool.ntp.org servers configured.


tests/tests_ntp_provider7.yml
```diff
@@ -0,0 +1,31 @@
...
+  vars:
+    timesync_ntp_servers:
+      - hostname: 172.16.123.1
+    timesync_peer_server:
+      - hostname: 127.0.0.1
+    timesync_ntp_provider: chrony
+  roles:
+    - linux-system-roles.timesync
+
+  pre_tasks:
```
pcahyna: what's the point of those three tasks?

tests/tests_ntp_provider7.yml
```diff
@@ -0,0 +1,31 @@
...
+    - name: Remove chrony
+      package: name=chrony state=absent
+
+  post_tasks:
+    - name: Wait for services to start
+      wait_for:
+        timeout: 2
+
+    - name: Check chronyd service
+      shell: chronyc -n tracking
```
pcahyna: please use the command module instead of shell if possible.

tests/tests_ntp_provider7.yml
pcahyna: I see:  https://github.com/linux-system-roles/timesync/blob/924650d0cd4117f73a7f0413ab745a8632bc5cec/tests/tests_ntp_provider6.yml#L9

tests/tests_ntp_provider7.yml
hairmare: TBH `tests_ntp_provider7.yml` is a simple copy of `tests_ntp_provider4.yml` with the additional param. Do you want me to change to command there as well? 

I'm not even sure where these tests get executed :/

pcahyna: No, if it is a copy, we will change them all together.

pcahyna:
@mlichvar could you please check whether the functionality and the interface makes sense? TIA

pcahyna:
[citest commit:4ae7b9491b2d41598d6dc596a244f546c4adc1f4]

hairmare:
The tests  found a smallish nag: "inline if-expression on line 21 evaluated to false and no else section was defined" which I fixed by adding an else section and repushing.

templates/ntp.conf.j2
```diff
@@ -7,14 +7,20 @@ value['hostname'] }}{{
...
 driftfile /var/lib/ntp/drift
 
 # Permit time synchronization with our time source, but do not
 # permit the source to query or modify the service on this system.
-restrict default nomodify notrap nopeer noquery
+restrict default nomodify notrap {{ 'nopeer' if timesync_ntp_peers|length == 0 else '' }} noquery
 {% if timesync_ntp_version.stdout is version('4.2.6', '<') %}
-restrict -6 default nomodify notrap nopeer noquery
+restrict -6 default nomodify notrap {{ 'nopeer' if timesync_ntp_peers|length == 0 else '' }} noquery
 {% endif %}
```
mlichvar: It's not clear to me why this is needed. As I understand it, `nopeer` applies only to unauthenticated packets when authentication is disabled with `disable auth`.

README.md
```diff
@@ -35,6 +35,14 @@ timesync_ntp_servers:
                                 # of the hostname is a separate NTP server
                                 # (default no)
 
+# List of NTP peers
+timesync_ntp_peers:
```
mlichvar: We would probably want to keep the API consistent with pool, so instead of adding a new variable for peers, there should just be a new flag in ntp_servers

mlichvar:
If the API used a "peer" flag instead of adding and it didn't change the default restrictions in ntp.conf, I think it could work.

But I'm not sure if we actually want to support the symmetric mode. It's less secure than the client/server mode and if we don't support ephemeral associations (which would require distribution of keys between the hosts), there is no advantage over the client/server mode.

hairmare:
I refactored to use a peer flag instead of a separate variable to configure peers and also removed the nopeer argument to restrict.

I've been trying to grok `man 5 ntp_auth` and [draft-ietf-ntp-bcp-08](https://tools.ietf.org/id/draft-ietf-ntp-bcp-08.html#rfc.section.5). To my understanding symmetric mode isn't an issue on a trusted network with trusted peers only. In other cases the ntp peer should absolutely be keyed. 
As far as I can tell the draft BCP recommends setting up PSK as well (and it recommends not to use autokey). Maybe I'm not understanding this correctly but my interpretation of all the docs I've come across so far would mandate setting up keys if you need secure ntp peering. I can't seem to figure out why/if this should not also apply to a persistent client mode association.

templates/ntp.conf.j2
hairmare: I removed this in 17a3a7d6c112c2a686edf59372168b0121595143, I checked on my production and it was indeed wrong.

README.md
hairmare: Refactored in 17a3a7d6c112c2a686edf59372168b0121595143

mlichvar:
Can you please explain why do you need this? I.e. what difference does it make in your environment to specify the servers as `peer` instead of `server`? My understanding is that unless you disable authentication or distribute keys (which the role currently doesn't support), there shouldn't be any.

hairmare:
I want to ensure that my "stratum 2" servers always peer with each other while still associating with configured pool servers. I originally had my internal ntp cluster setup in a client/server & server/client config for local redundancy, at some times the peer members would drift if they didn't consider each other candidates. Using peer seemed to have helped ensure that the local cluster is always in sync while still considering some external pool servers.

mlichvar:
At least in ntpd and chronyd, there is no difference in the source selection between peers and servers. The same rules apply to both. If you saw a difference it was probably for a different reason.  For keeping time in isolated networks, or when the connection is down, there is the orphan mode, which is not related to the peer/server mode.

Security might not be the only issue here. Adding the peer option to this role would actually change its purpose.  I think it was originally intended for clients only. The fact that ntpd configured by this role operates as a server is due to missing a simple way to disable the server.

@pcahyna, what do you think?

richm:
Can we close this PR?

richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

richm:
[citest commit:17a3a7d6c112c2a686edf59372168b0121595143]

richm:
please rebase this on top of the latest master branch

mlichvar:
I'm still not convinced this is a good idea. Symmetric mode has multiple disadvantages over the client/server mode. It will likely not even be supported in the next NTP protocol (NTPv5).

If someone think they really need it, I'd suggest to use the custom_settings variable for that.

richm:
> I'm still not convinced this is a good idea. Symmetric mode has multiple disadvantages over the client/server mode. It will likely not even be supported in the next NTP protocol (NTPv5).
> 
> If someone think they really need it, I'd suggest to use the custom_settings variable for that.

ok - then I'm closing this PR

PR #47, opened 2019-07-25
Enable line-length check. Fix found issues.

vcrhonek:


pcahyna:
LGTM, please merge when appropriate.

pcahyna:
[citest bad]

tasks/main.yml
```diff
@@ -27,7 +27,9 @@
 
 - name: Select NTP provider
   set_fact:
-    timesync_ntp_provider: "{{ timesync_ntp_provider_os_default if timesync_ntp_provider_current == '' else timesync_ntp_provider_current }}"
+    timesync_ntp_provider: "{{
+        timesync_ntp_provider_os_default if timesync_ntp_provider_current == ''
+        else timesync_ntp_provider_current }}"
```
pcahyna: One could use the default filter with the second parameter instead of this long if expression: https://jinja.palletsprojects.com/en/2.9.x/templates/#default

vars/main.yml
```diff
@@ -1 +1,4 @@
-timesync_ntp_provider_os_default: "{{ 'ntp' if ansible_distribution in ['RedHat', 'CentOS'] and ansible_distribution_version is version('7.0', '<') else 'chrony' }}"
+timesync_ntp_provider_os_default: "{{
+    'ntp' if ansible_distribution in ['RedHat', 'CentOS'] and
+        ansible_distribution_version is version('7.0', '<')
```
pcahyna: I think it would be better to move this variable to the distribution-specific vars files instead of this conditional.

tasks/main.yml
```diff
@@ -1,5 +1,11 @@
 ---
 
+- name: Gather OS specific variables
+  include_vars: "{{ item }}"
+  with_first_found:
+    - "{{ ansible_distribution }}-{{ ansible_distribution_major_version}}.yml"
+    - default.yml
```
pcahyna: https://github.com/linux-system-roles/template/pull/5/commits/0e1144ba1d0bdb972745b84912e18919742078cb#diff-2444ad0870f91f17ca6c2a5e96b26823R7

tasks/main.yml
```diff
@@ -27,7 +33,9 @@
 
 - name: Select NTP provider
   set_fact:
-    timesync_ntp_provider: "{{ timesync_ntp_provider_os_default if timesync_ntp_provider_current == '' else timesync_ntp_provider_current }}"
+    timesync_ntp_provider: "{{
+        timesync_ntp_provider_current|
+        default(timesync_ntp_provider_os_default, true)}}"
```
pcahyna: please add space before the braces. https://github.com/oasis-roles/meta_standards/blob/master/README.md#jinja2-syntax

tasks/main.yml
vcrhonek: Fixed.

tasks/main.yml
vcrhonek: Fixed.

mlichvar:
Looks good to me.

Issue #51, opened 2019-09-30
role fails in check mode if not having been executed in normal mode before


pcahyna:
Followup to #27 .

We fixed failures in check mode some time ago and now the role does not fail if executed in check mode after having been executed in normal mode. There is still a problem though when the role is executed in check mode on a fresh system (without having been applied in normal mode before), because it assumes that some packages are installed, and in check mode they do not get installed. This is the case described in https://github.com/ansible/molecule/issues/128#issuecomment-245009843 (note that the current Molecule setup does not test for it, as explained in that link).
The change submitted by @hhaniel in https://github.com/linux-system-roles/timesync/issues/27#issuecomment-472466223 should fix this case as well.
Test playbook:
```yaml
- name: Ensure that the role runs with default parameters
  hosts: all
  roles:
    - name: timesync
      vars:
        timesync_ntp_provider: ntp
```
error:

TASK [timesync : Get ntp version] **********************************************
fatal: [/home/pcahyna/linux-system-roles-testing/local/CentOS-7-x86_64-GenericCloud-1702.qcow2c]: FAILED! => {"changed": false, "cmd": ["rpm", "-q", "--qf", "%{version}", "ntp"], "delta": "0:00:00.033978", "end": "2019-09-30 14:53:22.430978", "msg": "non-zero return code", "rc": 1, "start": "2019-09-30 14:53:22.397000", "stderr": "", "stderr_lines": [], "stdout": "package ntp is not installed", "stdout_lines": ["package ntp is not installed"]}

PLAY RECAP *********************************************************************
/home/pcahyna/linux-system-roles-testing/local/CentOS-7-x86_64-GenericCloud-1702.qcow2c : ok=3    changed=1    unreachable=0    failed=1    skipped=9    rescued=0    ignored=0

pcahyna:
Note that the test uses the ntp provider, because chrony is installed in that image, therefore the problem does not show up with the default provider, at least not with the CentOS image I used.

PR #52, opened 2019-10-07
Fix check mode

vcrhonek:
Fix the role if executed in check mode without having been executed in normal mode before.
https://github.com/linux-system-roles/timesync/issues/51

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -65,6 +65,10 @@
     state: present
   when: timesync_mode != 1
 
+- name: Gather package facts
+  package_facts:
+    manager: auto
+
 - name: Run phc_ctl on PTP interface
   command: phc_ctl -q {{ timesync_ptp_domains[0].interfaces[0] }}
   register: timesync_phc_ctl_output
@@ -77,26 +81,6 @@
   set_fact:
     timesync_mode2_hwts: "{{ timesync_phc_ctl_output.rc == 0 }}"
   when: timesync_mode == 2
-
-- name: Get chrony version
-  command: rpm -q --qf '%{version}' chrony
-  args:
-    warn: false
-  register: timesync_chrony_version
-  changed_when: false
-  check_mode: false
-  when: timesync_mode != 2 and timesync_ntp_provider == 'chrony'
-
-- name: Get ntp version
-  command: rpm -q --qf '%{version}' ntp
-  args:
-    warn: false
-  register: timesync_ntp_version
-  changed_when: false
-  check_mode: false
-  when:
-    - timesync_mode != 2
-    - timesync_ntp_provider == 'ntp'
 
 - name: Generate chrony.conf file
   template:
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -92,6 +92,7 @@
   when:
     - timesync_mode != 2
     - timesync_ntp_provider == 'chrony'
+    - "'chrony' in ansible_facts.packages"
 
 - name: Generate chronyd sysconfig file
   template:
@@ -103,6 +104,7 @@
   when:
     - timesync_mode == 1
     - timesync_ntp_provider == 'chrony'
+    - "'chrony' in ansible_facts.packages"
 
 - name: Generate ntp.conf file
   template:
@@ -114,6 +116,7 @@
   when:
     - timesync_mode != 2
     - timesync_ntp_provider == 'ntp'
+    - "'ntp' in ansible_facts.packages"
 
 - name: Generate ntpd sysconfig file
   template:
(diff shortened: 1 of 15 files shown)


mlichvar:
The changes look good to me.

If it's not too much work, I think the switch to the package_facts would be better in a separate commit. Also, please consider using simpler commit subjects using only one sentence and moving the rest to the commit message, so the one-line git log looks nice.

The failure on RHEL8 looks like a selinux issue. I'll file a bug if it's something new.

pcahyna:
@mlichvar I believe it is https://bugzilla.redhat.com/show_bug.cgi?id=1668667 which should be fixed in the later snapshots (the rhel-x images).

mlichvar:
It seems there is another selinux issues that breaks the L2 transport and is not caught by our tests.
https://bugzilla.redhat.com/show_bug.cgi?id=1759214

I'll submit a PR to improve the test.

vcrhonek:
> The changes look good to me.

Thanks for review.

> If it's not too much work, I think the switch to the package_facts would be better in a separate
> commit.

Fixed. That particular commit also incorporates PR#49, because it touches that part of code anyway.

> Also, please consider using simpler commit subjects using only one sentence and moving
> the rest to the commit message, so the one-line git log looks nice.

Fixed.

tasks/main.yml
```diff
@@ -54,6 +54,11 @@
     state: present
   when: timesync_mode != 1
 
+- name: Gather package facts
+  package_facts:
+    manager: auto
+  register: _timesync_package_facts
```
pcahyna: You don't need to register the output from a fact module.

tasks/main.yml
```diff
@@ -68,24 +73,20 @@
...
 - name: Get chrony version
-  command: rpm -q --qf '%{version}' chrony
-  args:
-    warn: false
-  register: timesync_chrony_version
-  changed_when: false
-  check_mode: false
-  when: timesync_mode != 2 and timesync_ntp_provider == 'chrony'
+  set_fact:
+    timesync_chrony_version: "{{ _timesync_package_facts.ansible_facts.packages['chrony'][0].version|int }}"
```
pcahyna: just use ansible_facts.packages....
https://docs.ansible.com/ansible/latest/reference_appendices/common_return_values.html#ansible-facts

tasks/main.yml
vcrhonek: I did that intentionally because otherwise the facts could change, right? We discussed it elsewhere in the past. But I see that this might be a different situation and register is not necessary.

tasks/main.yml
```diff
@@ -186,53 +198,67 @@
...
+  when:
+    - timesync_mode != 1 or timesync_ntp_provider != 'ntp'
+    - "'ntp' in _timesync_package_facts.ansible_facts.packages"
   ignore_errors: true
 
 - name: Disable ntpdate
   service:
     name: ntpdate
     state: stopped
     enabled: false
```
pcahyna: It might just work in check mode, and even if it does not, ignore_errors below should make the "when" unneeded.

tasks/main.yml
```diff
@@ -68,24 +73,20 @@
...
 
 - name: Get chrony version
-  command: rpm -q --qf '%{version}' chrony
-  args:
-    warn: false
-  register: timesync_chrony_version
-  changed_when: false
-  check_mode: false
-  when: timesync_mode != 2 and timesync_ntp_provider == 'chrony'
+  set_fact:
```
pcahyna: Please do not set a fact, use vars/main.yml and lazy evaulation, or set it on a block similarly to a (let) construct in LISP (which would have the advantage of making the variable really local to the block).

vcrhonek: Fixed.

tasks/main.yml
vcrhonek: Correct, fixed.

tasks/main.yml
vcrhonek: Fixed.

tasks/main.yml
vcrhonek: Fixed.

mlichvar:
The changes look good to me, but I'm not familiar with the lazy evaluation feature of ansible.  @pcahyna, can you please review the latest changes?

Please squash commits that fix previous changes if possible. (Don't squash all commits.)
- "Remove unnecessary 'when' conditions if task ignores failed commands" looks like a fix for "Fix role in check mode"
- "Don't register package facts" looks like a fix for "Use package_facts to determine chrony/ntp version".

vcrhonek:
I just rebased the pull request atop of the current master, no other changes done.

pcahyna:
> The changes look good to me, but I'm not familiar with the lazy evaluation feature of ansible. 

@mlichvar , I now wrote a longer explanation of lazy evaluation, here: https://github.com/linux-system-roles/storage/pull/81#discussion_r406069895

PR #54, opened 2019-10-11
add timesync_ntp_custom_settings variable for free-form local configs

jontow:
Found this small addition useful in my local installation, thought it might be helpful to others.


mlichvar:
The role used to have some variables for adding settings directly to the generated config files, but IIRC they were removed in order to hide the implementation-specific details.

Later was added a variable for selecting the NTP provider, so it may no longer be a goal.

mcaubet:
I would agree with this change, in fact I was checking how would be possible to add extra options in chrony.conf which should apply a similar change to the one proposed here. We rely on some customized settings and with the current templates we are not able to deploy them. I will create a new merge request for chrony based on this pull request (for keeping coherency in case that this is merged).

richm:
[citest commit:ab66e75457880e7184a2d7b27d428f566ca7c40e]

tests/tests_ntp.yml
```diff
@@ -12,6 +12,8 @@
         iburst: yes
         minpoll: 4
         maxpoll: 6
+    timesync_ntp_custom_settings:
```
richm, line 15: Please update README.md with this parameter

jontow: Ok, README section in place!

tests/tests_ntp.yml
```diff
@@ -11,6 +11,8 @@
...
+      - "tinker panic 0"
```
richm, line 15: The test should check if this setting was applied correctly

richm:
@mlichvar @pcahyna This seems like a good feature - do we have a best practices for specifying provider specific settings?

tests/tests_ntp.yml
jontow: I'm not sure how to check this specific setting by inspecting runtime state, the closest I can come up with is looking for this line in the generated config file.  Is that enough, or are you thinking something more?

richm: > I'm not sure how to check this specific setting by inspecting runtime state, the closest I can come up with is looking for this line in the generated config file. Is that enough, or are you thinking something more?

I'll defer to @mlichvar about that - but IMO it is better to check the generated config file than to do nothing.

README.md
```diff
@@ -64,6 +64,11 @@ timesync_min_sources: 1
 # or enabled service will be configured. If no service is active or enabled, a
 # package specific to the system and its version will be selected.
 timesync_ntp_provider: chrony
+
+# List of custom settings passed through to ntp.conf.  If not defined,
+# setting is ignored.  Settings should be one per line (list item).
+timesync_ntp_custom_settings:
+  - "tinker panic 0"
```
richm, line 71: How idempotent is this? For example, if the user specifies
```yaml
timesync_ntp_custom_settings:
  - "tinker panic 0"
  - "a b c"
```
then sometime later specifies
```yaml
timesync_ntp_custom_settings:
  - "a    b    c"
  - "tinker panic 0"
```
Will that generate a change?  Does it matter if that generates a change?  For example, that will cause the ntp services to restart - is that ok?

richm: Do you want to allow the user to perform incremental changes?  For example, if the configuration already has
```yaml
timesync_ntp_custom_settings:
  - "tinker panic 0"
  - "a b c"
```
and the user wants to remove "a b c" - how to do that?  Or the user wants to remove "a b c" and add "b c a"?  For the kernel_settings role we had a similar requirement, so we documented it - https://linux-system-roles.github.io/documentation/incremental_settings.html

richm:
[citest]

tests/tests_ntp.yml
mlichvar: There might be a possibility to do that with `ntpq`, but it would need authentication to be set up. I think it's better to check the generated file.

mlichvar:
I'm not aware of any best practices for injecting configuration snippets like this. I think it would be good to warn the users that there are no guarantees a working config will not stop working with a newer version of the role if the template is changed for instance. We don't want to make any assumption about what is injected in the config.

richm:
[citest bad]

richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

richm:
I'll note that https://github.com/linux-system-roles/timesync/pull/80 did this for chrony settings and that PR includes a test to verify the custom settings - something similar should be done for the custom ntp settings

richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

richm:
please rebase to latest master branch

richm:
[citest pending]

richm:
closing due to inactivity

PR #55, opened 2020-01-13
Refactor role

vcrhonek:
Move constants to var files. Parametrize repeated tasks - package install,
service enable/disable, config/sysconfig file generation.

### DIFF
--- a/vars/main.yml
+++ b/vars/main.yml
@@ -1,4 +1,6 @@
 __timesync_chrony_version: "{{
-  ansible_facts.packages['chrony'][0].version|default('')|int }}"
+  (( ansible_facts.packages.chrony | default([]) )[0]
+    | default(dict()) )['version'] | default('0') }}"
 __timesync_ntp_version: "{{
-  ansible_facts.packages['ntp'][0].version|default('')|int }}"
+  (( ansible_facts.packages.ntp | default([]) )[0]
+    | default(dict()) )['version'] | default('0') }}"
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: python
 services:
   - docker
 install:
-  - pip install molecule docker
+  - pip install 'molecule<3' docker
 script:
   - molecule --version
   - ansible --version
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -1,25 +1,25 @@
 ---
 - name: restart chronyd
   service:
-    name: chronyd
+    name: "{{ timesync_chrony_service }}"
     state: restarted
 
 - name: restart ntpd
   service:
-    name: ntpd
+    name: "{{ timesync_ntp_service }}"
     state: restarted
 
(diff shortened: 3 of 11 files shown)


mlichvar:
The changes look good to me, but the tests seem to be failing (different tests on different systems). Is that expected?

vcrhonek:
Thanks. It seems that only 'tests_ntp_ptp' and 'tests_ntp' are failing. That is not expected, I'll take a look.

vcrhonek:
Fixed. It shows up that 'import_tasks' silently ignores 'notify', thus handlers were not triggered and that caused failures of the tests.

I've also updated handlers, because I forgot to use variables for the service names in original patch.

vcrhonek:
But I see that some tests are still failing, I'll investigate...

pcahyna:
Molecule is easy - fails on line length, but the other failures seem more serious.

vcrhonek:
[citest]

vcrhonek:
I've fixed the line length. Tests pass locally for me, I'm not able to reproduce fails (for example tests_ntp_ptp.yml on CentOS 6). It seems that the same issue is in PR #57.

pcahyna:
Thanks. Could it be because of Ansible version?

vcrhonek:
Yes, it seems that it is because of Ansible version.

When chrony is not installed, __timesync_chrony_version: "{{ ansible_facts.packages['chrony'][0].version|default('')|int }}" works as expected with ansible 2.9.4 (f30) but fails with ansible 2.7.10 (f28).

vcrhonek:
[citest]

vars/main.yml
```diff
@@ -1,4 +1,6 @@
 __timesync_chrony_version: "{{
-  ansible_facts.packages['chrony'][0].version|default('')|int }}"
+  (( ansible_facts.packages.chrony | default([]) )[0]
+    | default(dict()) )['version'] | default('0') }}"
 __timesync_ntp_version: "{{
-  ansible_facts.packages['ntp'][0].version|default('')|int }}"
+  (( ansible_facts.packages.ntp | default([]) )[0]
+    | default(dict()) )['version'] | default('0') }}"
```
richm, line 6: Is this a better way to fix this issue?  https://github.com/linux-system-roles/timesync/pull/67/commits/8772d38af1c03b5fa02549b7b73a5071c9fe2237

richm:
@vcrhonek do we still need this PR?  If so, it will have to be rebased and reworked slightly (e.g. the tox.ini changes have been made obsolete by tox-lsr plugin).

vcrhonek:
> @vcrhonek do we still need this PR? If so, it will have to be rebased and reworked slightly (e.g. the tox.ini changes have been made obsolete by tox-lsr plugin).

The intention of this PR was to simplify the role and make it more clear. I'm really not sure whether it succeeds in this. Personally, I wouldn't mind if this PR was closed.

PR #60, opened 2020-03-05
Fix version determination

vcrhonek:


pcahyna:
While #30944 does not indicate any intent to fix it, the changelog of Ansible 2.8 says that that this is fixed: https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_2.8.html#jinja-undefined-values and https://github.com/ansible/ansible/blob/stable-2.8/changelogs/CHANGELOG-v2.8.rst#id25 "jinja2 - accesses to attributes on an undefined value now return further undefined values rather than throwing an exception
jinja2 - accesses to keys/indices on an undefined value now return further undefined values rather than throwing an exception"
so when 2.7 stops being supported, we can get rid of this ugliness.

pcahyna:
here is a link to a solution of a similar problem in logging by @nhosoi. We should agree on a common standard. 
https://github.com/linux-system-roles/logging/pull/87/files#diff-ed5a08a2e61430dc18e2d3820cbd6f32R15

pcahyna:
While https://github.com/ansible/ansible/issues/30944 does not indicate any intent to fix it, the changelog of Ansible 2.8 says that that this is fixed: https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_2.8.html#jinja-undefined-values and https://github.com/ansible/ansible/blob/stable-2.8/changelogs/CHANGELOG-v2.8.rst#id25 "jinja2 - accesses to attributes on an undefined value now return further undefined values rather than throwing an exception
jinja2 - accesses to keys/indices on an undefined value now return further undefined values rather than throwing an exception"
so when 2.7 stops being supported, we can get rid of this ugliness.

vcrhonek:
> here is a link to a solution of a similar problem in logging by @nhosoi. We should agree on a common standard.
> https://github.com/linux-system-roles/logging/pull/87/files#diff-ed5a08a2e61430dc18e2d3820cbd6f32R15

Personally, I like the solution from logging better. I find it more readable.

pcahyna:
@vcrhonek it did not work though: https://github.com/linux-system-roles/logging/pull/91 https://github.com/linux-system-roles/logging/pull/87#discussion_r389212373

nhosoi:
> @vcrhonek it did not work though: [linux-system-roles/logging#91](https://github.com/linux-system-roles/logging/pull/91) [linux-system-roles/logging#87 (comment)](https://github.com/linux-system-roles/logging/pull/87#discussion_r389212373)

FYI, the following 2 work.
https://github.com/linux-system-roles/logging/pull/87/files#diff-ed5a08a2e61430dc18e2d3820cbd6f32R15
https://github.com/linux-system-roles/logging/pull/91/files#diff-ed5a08a2e61430dc18e2d3820cbd6f32R16

This selectattr filter does not work.
https://github.com/linux-system-roles/logging/pull/91/files#diff-ed5a08a2e61430dc18e2d3820cbd6f32L16

vars/main.yml
```diff
@@ -1,4 +1,6 @@
 __timesync_chrony_version: "{{
-  ansible_facts.packages['chrony'][0].version|default('')|int }}"
+  (( ansible_facts.packages.chrony | default([]) )[0]
+    | default(dict()) )['version'] | default('0') }}"
 __timesync_ntp_version: "{{
-  ansible_facts.packages['ntp'][0].version|default('')|int }}"
+  (( ansible_facts.packages.ntp | default([]) )[0]
+    | default(dict()) )['version'] | default('0') }}"
```
pcahyna: I think it can be shortened by using `d()` instead of `default()` and `{}` instead of `dict`. Still a bit ugly though and the logging solution may be better.

pcahyna:
[citest bad]

vcrhonek:
> @vcrhonek it did not work though: [linux-system-roles/logging#91](https://github.com/linux-system-roles/logging/pull/91) [linux-system-roles/logging#87 (comment)](https://github.com/linux-system-roles/logging/pull/87#discussion_r389212373)

Yes, I'm aware of that and I was referring to fixed variant from logging:

{{ ansible_facts.packages['chrony'][0].version
    if ansible_facts.packages | json_query('chrony')
    else '0' }}

pcahyna:
[citest]

pcahyna:
[citest]

pcahyna:
[citest]

richm:
@vcrhonek @pcahyna what about this solution?  https://github.com/linux-system-roles/timesync/pull/67/commits/8772d38af1c03b5fa02549b7b73a5071c9fe2237

pcahyna:
@richm looks better than using `json_query`, thanks! The same change should be then done to logging to make this uniform: https://github.com/linux-system-roles/logging/pull/91/files#diff-d841fad1aebbeefea3e6aa82439eea65

vcrhonek:
@richm @pcahyna I agree

richm:
> @richm looks better than using `json_query`, thanks! The same change should be then done to logging to make this uniform: https://github.com/linux-system-roles/logging/pull/91/files#diff-d841fad1aebbeefea3e6aa82439eea65

done - https://github.com/linux-system-roles/logging/pull/101

richm:
I would like to delete this PR in favor of https://github.com/linux-system-roles/timesync/pull/67 which fixes the version determination issue, among other things.  OK?

vcrhonek:
Sure, please go ahead @richm and close it. Thank you!

richm:
closing in favor of #67

PR #62, opened 2020-03-09
Addition of SLES and Ubuntu support

chiqomar:
*Why?*
We have a role in our private repository called `common` that ensure CIS compliance and other standard tasks for the servers on our system. While `geerlingguy/ansible-role-ntp` and `ericsysmin/ansible-role-chrony` are quite good, using them together was quite a hassle (we need to be able to manage both `chrony` and `ntpd` installs). This role solves the issue, with a minimal amount of issues. But, it was also terribly hard to follow, which is why I used @vcrhonek's branch as the base for my changes.

*This change addresses the need by*
We need to support both RHEL6 and RHEL7 servers primarily, but have just added support for SLES 12 and SLES 15 servers, as well as Ubuntu 16 and Ubuntu 18; the molecule playbook contains a reference to our SLES registration role, which is not public; the molecule tests will require registered instances. OpenSuSE has not been tested, as it is not necessary
I did some extensive refactoring here as well, so it is easier to follow and is broken up into task files somewhat reasonably; there is probably some more work to do to make it truly DRY, but it's a good start.
I also pinned the meta Ansible version to 2.9, as that is what I am using primarily, but we can do some compatibility checks to bring it back a few minor versions.


meta/main.yml
```diff
@@ -14,7 +14,7 @@ galaxy_info:
     - ptp
   company: Red Hat, Inc.
   license: MIT
-  min_ansible_version: 2.4
+  min_ansible_version: 2.9
```
pcahyna: Is there anything that requires Ansible 2.9? 2.7 and 2.8 are still supported, we need to support them.

poetry.lock
```diff
@@ -0,0 +1,1475 @@
+[[package]]
```
pcahyna: what does this file do?

ansible.cfg
```diff
@@ -0,0 +1,19 @@
+[defaults]
```
pcahyna: Do you really need a custom ansible.cfg to use the role? Why? Our roles should work with the default `ansible.cfg`

tasks/main.yml
```diff
@@ -1,235 +1,58 @@
 ---
-
 - name: Set version specific variables
   include_vars: "{{ item }}"
   with_first_found:
-    - "{{ ansible_distribution }}-{{ ansible_distribution_version }}.yml"
-    - "{{ ansible_distribution }}-{{ ansible_distribution_major_version }}.yml"
+    - "{{ ansible_distribution }}_{{ ansible_distribution_version }}.yml"
+    - "{{ ansible_distribution }}_{{ ansible_distribution_major_version }}.yml"
```
pcahyna: why this change?

vars/RedHat_6.yml
```diff
@@ -1,4 +1,4 @@
-timesync_ntp_provider_os_default: "ntp"
+timesync_ntp_provider_os_default: "chrony"
```
pcahyna: In RHEL 6 the default was ntpd (in the sense that it was installed by default, chrony was not, although it was available).

tasks/main.yml
```diff
@@ -1,235 +1,58 @@
...
 
 - name: Check if single PTP is needed
   set_fact:
-    timesync_mode: 2
+    timesync_mode: ptp
   when:
     - timesync_mode is not defined
-    - timesync_ntp_servers|length == 0
-    - timesync_ptp_domains|length == 1
+    - timesync_ntp_servers|length in [0,1]
```
pcahyna: is this correct? it is not equivalent to the previous version.

templates/timemaster.conf.j2
```diff
@@ -18,17 +18,17 @@ ptp4l_option udp_ttl {{ value['udp_ttl'] }}
...
-includefile /etc/ntp.conf
+includefile {{ timesync_ntp_conf_path }}
 
 [ptp4l.conf]
 
 [chronyd]
 path /usr/sbin/chronyd
 {% if timesync_ntp_provider == 'chrony' and __timesync_chrony_version is version('1.30', '<') %}
-options -u chrony
+options -u {{ 'chrony' if anisble_os_family != 'Debian' else '_chrony' }}
```
pcahyna: sorry, but this was not tested. (typo.)
I propose to make it a variable and put it into a distribution-specific var file, instead of adding conditionals.

pcahyna:
Thanks, this is is certainly very interesting, but I am afraid this PR is doing too much, many of the changes seem to be unneeded for the result that you want, esp. in commit 4cf4c53e7ecb1b0d800dc663284d99da2a189dae. E.g. the .gitignore file, a large binary .pyc file, whitespace changes that, although in many cases reasonable, do not have anything to do with the subject... The SLES tests now cause Travis to fail, is there a way to do equivalent tests on something publicly available, like Opensuse? (Just as we are using CentOS instead of RHEL in Molecule.)
The Molecule changes are a bit controversial, there is repeated boilerplate (molecule/*/molecule.yml) that will be hard to maintain (we are generally not using Molecule much because of this and relying on our test harness: https://github.com/linux-system-roles/test-harness). Some of the changes like .pre-commit-config.yaml are good candidates for https://github.com/linux-system-roles/template - they don't seem specific to this role. OTOH `.vscode/settings.json` is specific to your environment and should be removed.
Could you please split this commit into smaller logically related changes and keep only those that you need in this PR? By the way, we are following the code style in https://github.com/oasis-roles/meta_standards and generally the examples set by oasis-roles - although you will see that old code is not consistent with it, it should be used for new code.

meta/main.yml
chiqomar: I don't believe this will be a problem, I just am currently using 2.9, and didn't want to jump the gun on saying it works.

poetry.lock
chiqomar: This is the lock file for the actual packages; it's recommended to check it in to make sure the packages being used are exactly the same, ala `package.lock` for Node.js package management. This can be removed if not desired.

ansible.cfg
chiqomar: Sorry, taking this out; just need it for some of our environments.

tasks/main.yml
chiqomar: I prefer underscores over dashes, but it appears that dashes are more desirable in filenames. Changing this.

vars/RedHat_6.yml
chiqomar: Yup, meant to revert this. We just want to make chrony the default for our environment.

tasks/main.yml
chiqomar: my bad.

tasks/main.yml
chiqomar: Oh. Looks like I have underscores supported per here: https://github.com/oasis-roles/meta_standards#naming-things

pcahyna: Good point - I did not realize it applied to files as well. I will discuss this with OASIS colleagues, because I always had the impression that dashes are more idiomatic in file names, but maybe we will change this for consistency.
Anyway, I think that such stylistic changes do not belong to this PR. Please keep it focused to the topic of SLES and Ubuntu support.

poetry.lock
pcahyna: Packages being used for what? Tests? The role does not use any Python code - why should it need so many Python packages?

chiqomar:
I'm going to abandon this PR; too many things to do to work with this here; but I should be able to get back to this or similar by the time I'm done.

PR #67, opened 2020-03-31
add support for tox; fix yamllint issues

richm:


tests/tests_ntp.yml
```diff
@@ -1,4 +1,4 @@
-
+---
```
pcahyna, line 1: This should not be needed.
```yaml
  document-start: disable
```
is specified in the yamllint configuration.

tests/tests_ntp.yml
```diff
@@ -20,20 +20,20 @@
 
   tasks:
     - block:
-      - meta: flush_handlers
+        - meta: flush_handlers
```
pcahyna, line 23: :+1: for conforming to the style of Ansible docs: https://docs.ansible.com/ansible/latest/user_guide/playbooks_blocks.html

tests/tests_ntp.yml
richm: The error was this:
```
./tests/tests_ntp.yml
  1:1       error    too many blank lines (1 > 0)  (empty-lines)
```
That is, the file begins with a blank line.  I guess `document-start: disable` is meant to ignore when the file starts with the actual yaml instead of the `---`?  Because if the file looks like this:
```yaml
- name: Configure time synchronization with NTP servers
  hosts: all
...
```
(note: no blank line at the beginning of the file), then `yamllint` reports no errors - the `document-start: disable` directive ignores that the file doesn't start with `---`.

richm: It took only a few minutes to add `---` to the beginning of the files that reported this error, so I just went ahead and did it.

pcahyna: It is against the OASIS best practices: https://github.com/oasis-roles/meta_standards/blob/master/README.md#yaml-and-jinja2-syntax

pcahyna: why they reported an error? The template properly avoids it.
https://github.com/linux-system-roles/template/blob/5bac4c7337af36b31c3191e4cd3aa4ddff2277bd/.yamllint.yml#L14

richm: ok - latest commit removes all leading `---` from yaml files

pcahyna: the fact that you saw an error is still suspect

richm: > the fact that you saw an error is still suspect

No, the error makes sense to me.  Try it yourself.

pcahyna: Ah, you saw a different error. Yes, `document-start: disable` disables the check for `---` at the beginning, not for blank lines. Blank lines are unrelated to the leading `---` marker. Removing the blank lines was right, they always seemed silly to me, I don't know why some YAML files have them.

richm:
ok to merge?  should I squash?

richm:
[citest pending]

LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2017 Red Hat, Inc.
```
pcahyna, line 1: IMO LICENSE should not be synchronized from the template at all. it is a choice that you make at the initial iport of the role, and one can not change it later to conform to some "standard".

pcahyna:
> ok to merge? should I squash?

See my comment regarding LICENSE, otherwise it looks good. And yes, please squash.

LICENSE
richm: There was no `LICENSE` file previously - there was only a `COPYING` file.  I added the file `LICENSE` which is just a copy of `COPYING`, to make it consistent with the template.  I did not add a new license or change the existing license.  I note that these repos have only a `COPYING` file and no `LICENSE`: firewall linux-system-roles.github.io postfix selinux test-harness tuned - is this OK?  If so, I'll remove `LICENSE` from this PR, and update the auto-maintenance/sync-template.sh script to skip copying `LICENSE` if `COPYING` already exists.

richm: I removed `LICENSE` from the PR.

pcahyna: I believe sync-template should just avoid copying LICENSE. It does not sound right to add a license if there is none. The repos should have a license in some form, but if they don't, one should not add one as part of some automated process, easily missed among other changes. By doing this, we would effectively relicense the project.
In other words, adding a license file should be a conscious act of the maintainer, not of an automated script.

pcahyna:
In general, it would be probably better to have automated sync as one commit and manual fixups (like those done to appease yamllint) as separate commits. But don't bother separating them now.
(There were two commits, but I think they did not respect this division, IIRC the first contained both template sync and manual changes.)

LICENSE
richm: ok - my latest commit in PR https://github.com/linux-system-roles/auto-maintenance/pull/9 removes `LICENSE`

PR #72, opened 2020-04-20
[RFE] Checkmode and indempotency testing

i386x:
Initial proposal of how to test check mode and idempotency in linux system roles. Please, write your comments/issues/recommendations here.


tests/role_wrapper.yml
```diff
@@ -0,0 +1,65 @@
...
+- name: >
+    Import role {{ tests_role_path }} via fresh Ansible instance with check
+    mode enabled
+  command: >
+    ansible all -vv -i {{ tempinventory.path }} --check -m import_role
+      -a "name={{ tests_role_path }}"
+  delegate_to: localhost
+  when:
+    - tests_test_checkmode is defined
+    - tests_test_checkmode
```
richm, line 28: these two lines can be combined into `tests_test_checkmode | d(false)` - same with the other similar `when` clauses below

tests/role_wrapper.yml
```diff
@@ -0,0 +1,65 @@
...
+    that:
+      - ansible_output.stdout.find('| CHANGED => {\n') == -1
+    fail_msg: >
+      Idempotency test failed. Output: {{ ansible_output.stdout_lines }}
+  delegate_to: localhost
+  when:
+    - tests_test_idempotency is defined
+    - tests_test_idempotency
+
+- name: Remove the temporary file
```
richm, line 60: This should be moved to before `Check the idempotency` - if that fails, the temporary file will not be removed.  Unless there is some other mechanism by which you can ensure clean up actions are taken. Or, use `block` https://docs.ansible.com/ansible/latest/user_guide/playbooks_blocks.html#blocks-error-handling

i386x:
Initial proposal of how to use `role_wrapper.yml`:
* `role_wrapper.yml` supports two `bool` parameters: `tests_test_checkmode` and `tests_test_idempotency`. It is supposed that either `tests_test_checkmode` or `tests_test_idempotency` is set, but not both.

Example of usage in playbooks:
1. neither check mode nor idempotency testing is enabled:
   ```yaml
   - name: Plain test
     hosts: all

     tasks:
       - include_tasks: role_wrapper.yml
   ```
1. check mode testing is enabled:
   ```yaml
   - name: Check mode test
     hosts: all
     vars:
       tests_test_checkmode: yes

     tasks:
       - include_tasks: role_wrapper.yml
   ```
1. idempotency testing is enabled:
   ```yaml
   - name: Idempotency test
     hosts: all
     vars:
       tests_test_idempotency: yes

     tasks:
       - include_tasks: role_wrapper.yml
   ```
Proposal of how to convert tests:
1. Identify what to test.
1. Convert `tests_X` to `tests_X`, `tests_X_checkmode`, `tests_X_idempotency`.
   1. Wrap common code of the old `tests_X` to `__tests_X_tasks` (TODO: choose better name?)
   1. Include `__tests_X_tasks` to new `tests_X`, `tests_X_checkmode`, `tests_X_idempotency`, as proposed above.

tests/inventory.yaml.j2
```diff
@@ -1 +1,13 @@
-{{ { 'all': { 'hosts': hostvars } } | to_yaml }}
+{% set role_vars = vars.keys() | intersect((
+     lookup('file', '../defaults/main.yml', errors='ignore') | from_yaml
+   ).keys())
+-%}
+{{ {
+  'all': {
+    'hosts': hostvars | to_json | from_json | combine(dict(
```
richm, line 7: `to_json | from_json` ?

tests/inventory.yaml.j2
```diff
@@ -1 +1,13 @@
-{{ { 'all': { 'hosts': hostvars } } | to_yaml }}
+{% set role_vars = vars.keys() | intersect((
```
richm, line 1: should be named `inventory.yml.j2` - the system roles best practices is to use `.yml` instead of `.yaml`

tests/role_wrapper.yml
```diff
@@ -0,0 +1,82 @@
+- name: Create a temporary file for a static inventory
+  tempfile:
+    state: file
+    suffix: .inventory.yaml
```
richm, line 4: `.yml` not `.yaml`

tests/role_wrapper.yml
```diff
@@ -0,0 +1,82 @@
...
+  register: tempinventory
+  changed_when: false
+  delegate_to: localhost
+
+- name: Create a static inventory from hostvars
+  template:
+    src: inventory.yaml.j2
```
richm, line 11: `.yml` not `.yaml`

tests/inventory.yaml.j2
i386x: The explanation of this construct is in [commit](https://github.com/linux-system-roles/timesync/pull/72/commits/7c4d07db2a65c8a7d41943ad070a3fe637366c89) message:
> The `hostvars | to_json | from_json` construct aims to convert HostVars
and HostVarsVars instances to Python dictionaries so `combine` filter
will apply correctly (note that `dict(hostvars)` is just not enough
because it only makes a shallow conversion to dict which ends up with
`hostvars[host]` will be totally ignored by `combine`).

In short, I need to convert `hostvars` to `dict` deeply, but I not found the better solution then `to_json | from_json` yet.

richm:
lgtm - tyll? pcahyna?

tests/tests_default.yml
```diff
@@ -1,5 +1,8 @@
 - name: Ensure that the role runs with default parameters
   hosts: all
+  vars:
+    tests_role_path: "../../timesync"
```
tyll: `../../timesync` seems to be the git working copy here. IMHO it is wrong to assume that it has a certain name since developers can clone it/name it differently. Also, I guess this would not work when the role is installed via RPM. Not sure if I really understand what is going on, here.

tests/role_wrapper.yml
tyll: This should maybe be in a an `always` block to ensure it will still run if something else fails.

tests/role_wrapper.yml
tyll: Is this file stored on the developer machine or on the test VM? I guess the developer machine. Then it should be named more clearly and maybe even be a temporary file to avoid conflicts/irritation for developers.

tests/tests_ntp_provider1_checkmode.yml
```diff
@@ -0,0 +1,34 @@
+- name: Configure NTP with default provider
```
tyll, line 1: It seems that this file and the idempotency file are almost identical. IMHO it would be better to move the common parts to a file that is then re-used with the different settings. Or use one file that uses the file twice but with different variables (if possible).

tyll:
IMHO this is missing some documentation about the idea of this test structure to allow to understand what is going on. Is this supposed to be a specific solution for the timesync role or just an example that should be used for other roles as well?

tests/role_wrapper.yml
pcahyna: yes, developer machine/control host, see `delegate_to: localhost` below

tests/inventory.yaml.j2
pcahyna: What do you mean by "Python dictionaries" in Jinja2 context?

tests/role_wrapper.yml
i386x: It is stored on control node/developer machine as temporary file under `/tmp`.

tests/tests_default.yml
i386x: `tests_role_path` variable was discarded and replaced by `{{ '..' | realpath }}` in https://github.com/linux-system-roles/timesync/pull/72/commits/02caf146bacdb9018f241d5c69b3960584571c91. With this improvement, tests should be able to run upstream. To make them working also downstream, when the role is installed via RPM, more sophisticated method is needed.

tests/role_wrapper.yml
i386x: Fixed in https://github.com/linux-system-roles/timesync/pull/72/commits/9c8124ec7cbca33a2cc98c9bc6f2f842d4ba448b.

tests/role_wrapper.yml
tyll: sorry for the confusion, I missed that it is actually using the tempfile module

tests/inventory.yaml.j2
i386x: I Jinja2 context, I mean objects that are instances of `collections.abc.MutableMapping`.

tests/tests_default.yml
tyll: Using the original name like `linux-system-roles.timesync` and a corresponding symlink in the `roles/` dir would work. This is what the role currently does:
https://github.com/linux-system-roles/timesync/tree/master/tests/roles

Why is path needed, here?

The downstream test only needs to remove/not ship this symlink.

richm:
> IMHO this is missing some documentation about the idea of this test structure to allow to understand what is going on.

Yes, I'm still trying to understand.

> Is this supposed to be a specific solution for the timesync role or just an example that should be used for other roles as well?

This is just an example of what will be used in other roles.

richm:
There is an alternate proposal - https://github.com/linux-system-roles/certificate/compare/master...seocam:idempotence-example

tests/role_wrapper.yml
```diff
@@ -0,0 +1,105 @@
...
+  changed_when: false
+  delegate_to: localhost
+  when:
+    - tempinventory is defined
+    - tempinventory.path is defined
+
+- block:
+    - name: Import role {{ tests_role_name }}
+      import_role:
+        name: "{{ tests_role_name }}"
```
richm, line 45: What is this import_role for?  Is this just to verify that the given role can be imported?

i386x:
Closing this in favor of #103.

PR #63, opened 2020-03-12
Adding suse ubuntu support

chiqomar:
Using the master branch as the base, and adding openSUSE Leap 15 and 42 in for testing (along with variables for SLES and RedHat that I'm currently using).
The main changes here are

variables for almost everything that can be different from system to system
loading the default.yml first, so those are multiple values don't need to be loaded
Changes to the library file to support Ubuntu
Changes to the Dockerfile to support OpenSUSE Leap (enables systemd)
Addition of 4 more images for molecule testing (hope you've got the space...)

PR #76, opened 2020-07-30
Remove ignore_errors for Disable services tasks

pavel-z1:
Currently, we have errors, like these during execution role:
```
TASK [timesync : Disable ntpd] *************************************************************************************************************************************************************
fatal: [host1]: FAILED! => changed=false 
  msg: 'Could not find the requested service ntpd: host'
...ignoring
fatal: [host2]: FAILED! => changed=false 
  msg: 'Could not find the requested service ntpd: host'
...ignoring

TASK [timesync : Disable ntpdate] **********************************************************************************************************************************************************
fatal: [host1]: FAILED! => changed=false 
  msg: 'Could not find the requested service ntpdate: host'
...ignoring
fatal: [host2]: FAILED! => changed=false 
  msg: 'Could not find the requested service ntpdate: host'
...ignoring
```

To prevent errors was added additional service existing check .

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -1,3 +1,4 @@
+---
 - name: Set version specific variables
   include_vars: "{{ lookup('first_found', ffparams) }}"
   vars:
@@ -12,6 +13,22 @@
         - "default.yml"
       paths:
         - "{{ role_path }}/vars"
+
+- name: Get services list using platform specific services manager
+  include_tasks: "{{ lookup('first_found', __timesync_ff_params) }}"
+  vars:
+    __timesync_ff_params:
+      files:
+        - "{{ ansible_facts['distribution'] ~ '_' ~
+              ansible_facts['distribution_major_version'] }}.yml"
+        - "default.yml"
+      paths:
+        - "{{ role_path }}/tasks/setup"
+
+- name: Check that variable 'timesync_services' is defined
+  assert:
+    that: timesync_services is defined
+    fail_msg: "Variable 'timesync_services' is not defined"
 
 - name: Check if only NTP is needed
   set_fact:
@@ -196,54 +213,59 @@
     name: chronyd
     state: stopped
     enabled: false
-  when: timesync_mode != 1 or timesync_ntp_provider != 'chrony'
-  ignore_errors: true
+  when:
+    - timesync_mode != 1 or timesync_ntp_provider != 'chrony'
+    - "'chronyd' in timesync_services"
 
 - name: Disable ntpd
   service:
     name: ntpd
     state: stopped
     enabled: false
-  when: timesync_mode != 1 or timesync_ntp_provider != 'ntp'
-  ignore_errors: true
+  when:
+    - timesync_mode != 1 or timesync_ntp_provider != 'ntp'
+    - "'ntpd' in timesync_services"
 
 - name: Disable ntpdate
   service:
     name: ntpdate
     state: stopped
     enabled: false
-  ignore_errors: true
+  when: "'ntpdate' in timesync_services"
 
 - name: Disable sntp
   service:
     name: sntp
     state: stopped
     enabled: false
-  ignore_errors: true
+  when: "'sntp' in timesync_services"
 
 - name: Disable ptp4l
   service:
     name: ptp4l
     state: stopped
     enabled: false
-  when: timesync_mode != 2
-  ignore_errors: true
+  when:
+    - timesync_mode != 2
+    - "'ptp4l' in timesync_services"
 
 - name: Disable phc2sys
   service:
     name: phc2sys
     state: stopped
     enabled: false
-  when: timesync_mode != 2 or not timesync_mode2_hwts
-  ignore_errors: true
+  when:
+    - timesync_mode != 2 or not timesync_mode2_hwts
+    - "'phc2sys' in timesync_services"
 
 - name: Disable timemaster
   service:
     name: timemaster
     state: stopped
     enabled: false
-  when: timesync_mode != 3
-  ignore_errors: true
+  when:
+    - timesync_mode != 3
+    - "'timemaster' in timesync_services"
 
 - name: Enable chronyd
   service:
--- a/tasks/setup/CentOS_6.yml
+++ b/tasks/setup/CentOS_6.yml
@@ -1,0 +1,15 @@
+---
+- name: Set variable with services List for CentOS 6
+  command: chkconfig --list  # noqa 303
+  register: __timesync_register_services
+  check_mode: no
+  changed_when: false
+
+- name: Set variable `timesync_services` with filtered uniq service names
+  set_fact:
+    timesync_services: |
+      {{ __timesync_register_services.stdout_lines |
+          map('regex_replace', '[ \t].*$', '') |
+          unique |
+          list }}
+  when: __timesync_register_services is defined
--- a/tasks/setup/default.yml
+++ b/tasks/setup/default.yml
@@ -1,0 +1,16 @@
+---
+- name: Set variable with services List for RedHat 7 and above systems
+  command: systemctl list-unit-files --no-legend *.service  # noqa 303
+  register: __timesync_register_services
+  check_mode: no
+  changed_when: false
+
+- name: Set variable `timesync_services` with filtered uniq service names
+  set_fact:
+    timesync_services: |
+      {{ __timesync_register_services.stdout_lines |
+          map('regex_replace', '[.]service.*$', '') |
+          map('regex_replace', '@$', '') |
+          unique |
+          list }}
+  when: __timesync_register_services is defined
--- a/tasks/setup/RedHat_6.yml
+++ b/tasks/setup/RedHat_6.yml
@@ -1,0 +1,15 @@
+---
+- name: Set variable with services List for RedHat 6
+  command: chkconfig --list  # noqa 303
+  register: __timesync_register_services
+  check_mode: no
+  changed_when: false
+
+- name: Set variable `timesync_services` with filtered uniq service names
+  set_fact:
+    timesync_services: |
+      {{ __timesync_register_services.stdout_lines |
+          map('regex_replace', '[ \t].*$', '') |
+          unique |
+          list }}
+  when: __timesync_register_services is defined


richm:
In general it is not a good idea to have so much platform and provider specific tasks with lots of `when` statements in the main tasks file.  Instead, use a per-platform or per-provider setup file e.g. https://github.com/linux-system-roles/cockpit/blob/master/tasks/main.yml#L10 to isolate the provider or platform specific tasks in their own files.

pavel-z1:
Hi @richm ,
I've updated logic, as you asked.

tasks/main.yml
```diff
@@ -13,6 +14,16 @@
       paths:
         - "{{ role_path }}/vars"
 
+- name: Get servies list using platform specific services manager
+  include_tasks: "{{ item }}"
+  with_first_found:
+    - "services-el{{ ansible_facts['distribution_major_version'] }}.yml"
+
+- name: Check that variable 'services' is defined
```
richm, line 22: Ansible variables created by `register` are global so must be prefixed by the role name - so use `timesync_services`

tasks/main.yml
```diff
@@ -13,6 +14,16 @@
       paths:
         - "{{ role_path }}/vars"
 
+- name: Get servies list using platform specific services manager
+  include_tasks: "{{ item }}"
+  with_first_found:
+    - "services-el{{ ansible_facts['distribution_major_version'] }}.yml"
```
richm, line 20:
```yaml
- name: Get services list using platform specific services manager
  include_tasks: "{{ lookup('first_found', __timesync_ff_params) }}"
  vars:
    __timesync_ff_params:
      files:
        - "{{ ansible_facts['distribution'] }}_{{ ansible_facts['distribution_major_version'] }}.yml"
        - "default.yml"
      paths:
        - "{{ role_path }}/tasks/setup"
```
This is based on https://github.com/oasis-roles/meta_standards#platform-specific-tasks
Note that you have to specify both the distribution and version - otherwise on e.g. Fedora `"services-el{{ ansible_facts['distribution_major_version'] }}.yml"` will evaluate to `services-el32.yml` which is incorrect.

With the above, you'll need to create a new directory `tasks/setup/` which will contain 3 files - `CentOS-6.yml`, `RedHat-6.yml`, and `default.yml` - the first two files will use essentially what you have in services-el6.yml , and `default.yml` will contain essentially what you have in services-el7.yml.

Also note that the register variable is global, so must be prefixed by the role name - so `timesync_services`

tasks/main.yml
```diff
@@ -196,54 +207,59 @@
     name: chronyd
     state: stopped
     enabled: false
-  when: timesync_mode != 1 or timesync_ntp_provider != 'chrony'
-  ignore_errors: true
+  when:
+    - timesync_mode != 1 or timesync_ntp_provider != 'chrony'
+    - "'chronyd' in services"
```
richm, line 212: here and elsewhere the variable should be named `timesync_services`

pavel-z1:
Hi @richm 
No problem. Code updated.

richm:
The `setup/` directory should be under `tasks/` not at the top level.
My other concern with this change is that it will break support for platforms other than centos/rhel/fedora.  Unfortunately the `service` module does not have a platform independent way to get a list of services.
@mlichvar @pcahyna ptal

pavel-z1:
Accident mistake with setup dir path.
Fixed.

pavel-z1:
Regarding supported platforms.
This role already limited to run only on RedHad Enterprise Linux 6,7,8 - https://github.com/linux-system-roles/timesync/blob/master/meta/main.yml:
```
  platforms:
    - name: Fedora
      versions:
        - 28
        - 29
        - 30
    - name: EL
      versions:
        - 6
        - 7
        - 8
```
Therefore, no any new limits will be after apply this change.

tasks/main.yml
```diff
@@ -13,6 +14,21 @@
       paths:
         - "{{ role_path }}/vars"
 
+- name: Get servies list using platform specific services manager
```
richm, line 17: `servies` -> `services`

tasks/main.yml
```diff
@@ -13,6 +14,21 @@
...
+  include_tasks: "{{ lookup('first_found', __timesync_ff_params) }}"
+  vars:
+    __timesync_ff_params:
+      files:
+        - "{{ ansible_facts['distribution'] }}_{{ ansible_facts['distribution_major_version'] }}.yml"
```
richm, line 22: line too long - yamllint error - https://travis-ci.org/github/linux-system-roles/timesync/jobs/714473558 - use
```
         - "{{ ansible_facts['distribution'] ~ '_' ~
                ansible_facts['distribution_major_version'] }}.yml"
```

tasks/main.yml
```diff
@@ -13,6 +14,21 @@
...
+        - "default.yml"
+      paths:
+        - "{{ role_path }}/tasks/setup"
+
+- name: Check that variable 'timesync_services' is defined
+  assert:
+    that: timesync_services is defined
+    fail_msg: "The list of services was not received. Check variable 'timesync_services'"
```
richm, line 30: line too long - yamllint error - https://travis-ci.org/github/linux-system-roles/timesync/jobs/714473558 - use
```
     fail_msg: >
       "The list of services was not received. Check variable 'timesync_services'"
```

tasks/setup/default.yml
```diff
@@ -0,0 +1,5 @@
+---
+- name: Set variable with services List for RedHat 7 and above systems
```
richm, line 2: yamllint is complaining about trailing spaces - I think this line has trailing spaces

richm:
> This role already limited to run only on RedHad Enterprise Linux 6,7,8 - https://github.com/linux-system-roles/timesync/blob/master/meta/main.yml:

ok.  If this PR is approved, the author of https://github.com/linux-system-roles/timesync/pull/63 will need to rebase

tasks/setup/default.yml
pavel-z1: Fixed

tasks/main.yml
pavel-z1: Fixed

tasks/main.yml
pavel-z1: Done

tasks/main.yml
pavel-z1: Fixed

richm:
ci fails - https://travis-ci.org/github/linux-system-roles/timesync/jobs/714497208 - ansible lint wants to use the `systemd` module instead of `command: systemctl --all --type service`
but since this is required in order to perform this task, tell ansible lint to ignore this error:
```
    command: systemctl --all --type service  # noqa 303
```
https://docs.ansible.com/ansible-lint/rules/rules.html#excluding-rules

pavel-z1:
Added `# noqa 303`

richm:
lgtm - @mlichvar ?

mlichvar:
I'm not sure about the `systemctl` command. We don't want the service name to match some random description of an unrelated service. The `list-unit-files` command should print all services on disk. Maybe something like this would work better?
```
systemctl list-unit-files --no-legend *.service
```

mlichvar:
On RHEL6, the command should probably be `chkconfig --list` instead of `initctl list`.

But I think there is a different issue with using the whole output as one string. For example, if the `ntpdate` service is installed, but `ntpd` is not, `'ntpd' in timesync_services` will still match, right? It would be better to have a proper module returning a list of services, or at least process the output somehow to make it friendly to the string matching.

pavel-z1:
Ok
I think that such regex will be enough:
```
- name: Disable ntpd
  service:
    name: ntpd
    state: stopped
    enabled: false
  when:
    - timesync_mode != 1 or timesync_ntp_provider != 'ntp'
    - timesync_services is regex("ntpd(\.service|\s).*")
```

pavel-z1:
Check return error for previous change:
```
 fatal: [centos-6]: FAILED! => {"msg": "The conditional check 'timesync_services is regex('chronyd(\\.service|\\s).*')' failed. The error was: Unexpected templating type error occurred on ({% if timesync_services is regex('chronyd(\\.service|\\s).*') %} True {% else %} False {% endif %}): expected string or bytes-like object\n\nThe error appears to be in '/home/travis/build/linux-system-roles/timesync/tasks/main.yml': line 211, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Disable chronyd\n  ^ here\n"}
```

Therefore, I've updated check to filter regex_search:
```
- name: Disable ntpd
  service:
    name: ntpd
    state: stopped
    enabled: false
  when:
    - timesync_mode != 1 or timesync_ntp_provider != 'ntp'
    - timesync_services.stdout | regex_search("^ntpd(\.service|\s).*", multiline=True, ignorecase=False)
```

mlichvar:
Thanks for the update. It seems a bit hackish to me. I'll let others decide if that's acceptable.

pavel-z1:
Thank you for review.
I've added `check_mode: no` to the systemctl and chkconfig commands to prevent errors with empty variable during testing in ansible check mode.

tasks/setup/default.yml
```diff
@@ -0,0 +1,6 @@
...
+  command: systemctl list-unit-files --no-legend *.service  # noqa 303
+  register: timesync_services
+  check_mode: no
+  changed_when: false
```
richm, line 6: change the register variable to `__timesync_register_services` and add another task like this:
```yaml
    - set_fact:
        timesync_services: |
          {{ __timesync_register_services.stdout_lines |
             map('regex_replace', '[.]service.*$', '') |
             map('regex_replace', '@$', '') |
             unique |
             list }}
```
then you can do service tests like this:
```
  ...
  when:
    - timesync_mode != 3
    - "timemaster" in timesync_services
```
not sure about how to parse the output of `chkconfig --list` - I suspect it is different than the output of systemd - but the idea is to generate a `list` type variable which is a unique list of service names

pavel-z1: Done

tasks/setup/CentOS_6.yml
```diff
@@ -0,0 +1,16 @@
...
+  changed_when: false
+
+- name: Set variable `timesync_services` with filtered uniq service names
+  set_fact:
+    timesync_services: |
+      {{ __timesync_register_services.stdout_lines |
+          map('regex_replace', '[.]service.*$', '') |
+          map('regex_replace', '@$', '') |
+          unique |
+          list }}
```
richm, line 15: `chkconfig --list` output looks like this:
```
# chkconfig --list
abrt-ccpp      	0:off	1:off	2:off	3:on	4:off	5:on	6:off
abrtd          	0:off	1:off	2:off	3:on	4:off	5:on	6:off
acpid          	0:off	1:off	2:on	3:on	4:on	5:on	6:off
anamon         	0:off	1:off	2:off	3:off	4:off	5:off	6:off
```
Note that there may be a combination of space and tab characters after the service name.  Try something like this to parse it into a list variable:
```

- name: Set variable `timesync_services` with filtered uniq service names
  set_fact:
    timesync_services: |
      {{ __timesync_register_services.stdout_lines |
          map('regex_replace', '[ \t].*$', '') |
          unique |
          list }}
```
that is - delete from the first space or tab to the end of the line.  This should give you a list like
abrt-ccpp
abrtd
acpid
anamon

pavel-z1: Guys, this will never will be ideal.
We should stop constant rebase like this. )

If you agree that we need this change please accept Pull Request.

I have checked `chkconfig --list` after service name it always return whitespace on CentOS 6 and CentOS 7.
I don't see requirement here to add tab handling.

richm: > Guys, this will never will be ideal.

But it should work and not cause errors.  The current commit https://github.com/linux-system-roles/timesync/pull/76/files#diff-d65d631b9c97cb4867ddb320fbf11cd7R11 will not work on EL6 and will cause errors.  So it must be fixed, one way or another.

> I have checked chkconfig --list after service name it always return whitespace on CentOS 6 and CentOS 7.

I see tabs on some of the services in my output.  Which is why I suggested to add the tabs.  Is it a problem to have a regex that looks for either spaces or tabs?

I appreciate the fact that this PR has taken an inordinately long time to review, and a lot of review comments and change requests (now 1 week and many changes requested), and I thank you for your patience.  I hope you appreciate that we try to have high standards for the benefit of our customers.

pavel-z1: Hi Richard
Now I understand.
Didn't say that in some cases there are problems on CentOS 6 with regex. Replaced regex.

richm:
[citest]

PR #71, opened 2020-04-16
Fix iburst and add new options

mlichvar:
The first patch fixes the issue #70. The second patchs adds the prefer option as requested in issue #61 and also the trust option.

PR #83, opened 2020-09-09
Add NTS support

mlichvar:
This adds a new flag for NTP sources to enable the nts option in chrony. When the ntpd provider is selected (which doesn't support NTS), it adds the noselect flag instead to avoid using that source without authentication. It would probably be better to fail the task when the configuration cannot be supported, but I didn't find a way to trigger an error from the template.


richm (approved):
lgtm

richm:
@mlichvar ok to merge?

mlichvar:
chrony with NTS support is not available in most distros yet (including RHEL), so I think we should wait a bit to avoid confusing users with an option that doesn't work.

richm:
[citest]

mlichvar:
I reworked the ntp support to add "disable ntp" when the nts option is used to effectively disable all NTP sources, not just the one that had the option to make the behavior more consistent with chrony.

richm:
Is there some way to test the nts support?

mlichvar:
We can check if the "nts" option is present in the config. But this would need to be limited to hosts that have chrony >= 4.0

mlichvar:
Suggestions on how to do that are welcome.

richm:
> Suggestions on how to do that are welcome.

My only concern is that the folks that need to give QE approval of this feature will be able to verify the feature is working.

richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

mlichvar:
Is there anything else needed for this PR to be merged?

richm:
> Is there anything else needed for this PR to be merged?

I just merged the EL 9 support PR.  So please rebase on top of the latest master.

Also, it looks like this PR has some overlap with https://github.com/linux-system-roles/timesync/pull/71?  Which one should merge first - this one or #71?

mlichvar:
#71 should go first. This one has just one commit, but there is a conflict, so I included the others. Please let me know if you would prefer a rebase after #71 is merged. Thanks.

mlichvar:
Changed the chrony.conf template to add `ntsdumpdir` with chrony >= 4.0 for better NTS support.

richm (approved):
lgtm

PR #87, opened 2020-10-15
Fix package versions

mlichvar:
Don't convert the version of chrony and ntp to int to not lose the
minor number after the dot.


richm (approved):
lgtm

PR #88, opened 2020-10-15
Add support for HW timestamping with NTP

mlichvar:
This PR adds options to enable NTP HW timestamping on specified interfaces, enable the interleaved mode and sample filtering in order to enable highly accurate and stable synchronization in local network.

### DIFF
--- a/templates/chrony.conf.j2
+++ b/templates/chrony.conf.j2
@@ -26,6 +26,17 @@ makestep {{ timesync_step_threshold if timesync_step_threshold > 0.0 else '1.0' 
 # Enable kernel synchronization of the real-time clock (RTC).
 rtcsync
 
+{% if __timesync_chrony_version is version('3.0', '>=') and
+        timesync_ntp_hwts_interfaces|length > 0 %}
+# Enable hardware timestamping.
+{% for interface in timesync_ntp_hwts_interfaces %}
+hwtimestamp {{ interface }}{% if __timesync_chrony_version is version('3.1', '>=') %}
+ minpoll {{ (timesync_ntp_servers | rejectattr('minpoll', 'undefined') |
+             map(attribute='minpoll') | list + [0]) | min }}{% endif %}
+
+{% endfor %}
+
+{% endif %}
 {% if timesync_min_sources > 1 %}
 # Increase the minimum number of selectable sources required to adjust
 # the system clock.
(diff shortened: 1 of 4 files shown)


richm:
is it possible to add a test for HW timestamping?

mlichvar:
It is possible, but we need machines with NICs that support HW timestamping support and/or restrict the test to run only there.

richm:
> It is possible, but we need machines with NICs that support HW timestamping support and/or restrict the test to run only there.

Ok.  I'm assuming chrony QE has such machines and will do this testing.  Then for upstream can we have a test that just checks the generated config file has the option in it?

mlichvar:
Yes, we can do that. Maybe we should modify the other tests to do that too. Don't try to test the actual functionality, only make sure the configuration file is as expected.

richm:
> Yes, we can do that. Maybe we should modify the other tests to do that too. Don't try to test the actual functionality, only make sure the configuration file is as expected.

Well, if you already have tests that test the actual functionality, I wouldn't change them.

richm:
if you rebase on top of the latest master, you will get github action ci testing

mlichvar:
Tests added for each new option.

richm:
There are some yamllint line-length errors - otherwise, lgtm

mlichvar:
Fixed line-length warnings and added minpoll setting to the hwtimestamp directive from specified minpolls of NTP sources.

templates/chrony.conf.j2
```diff
@@ -21,6 +26,17 @@ makestep {{ timesync_step_threshold if timesync_step_threshold > 0.0 else '1.0'
 # Enable kernel synchronization of the real-time clock (RTC).
 rtcsync
 
+{% if __timesync_chrony_version is version('3.0', '>=') and
+        timesync_ntp_hwts_interfaces|length > 0 %}
+# Enable hardware timestamping.
+{% for interface in timesync_ntp_hwts_interfaces %}
+hwtimestamp {{ interface }}{% if __timesync_chrony_version is version('3.1', '>=') %}
+ minpoll {{ [ timesync_ntp_servers | map(attribute='minpoll', default=6) | min, 0] | min
```
richm, line 34: https://fedorapeople.org/groups/linuxsystemroles/logs/linux-system-roles-timesync-pull-linux-system-roles_timesync-88-ebd5733-rhel-8-20201215-154341/artifacts/ansible.log
```
TASK [linux-system-roles.timesync : Generate chrony.conf file] *****************
task path: /tmp/tmp0blwj5i4/tasks/main.yml:106
fatal: [/cache/rhel-8.qcow2]: FAILED! => {"changed": false, "msg": "FilterArgumentError: Unexpected keyword argument 'default'"}
```
I guess `map` in jinja2.7 does not support the `default` parameter - https://github.com/pallets/jinja/blob/2.7.3/jinja2/filters.py#L798

https://github.com/pallets/jinja/blob/2.7.3/docs/templates.rst

mlichvar: Ok. I'm trying to replace it with the rejectattr filter.

PR #89, opened 2020-10-19
Add option to configure max distance 

Thulium-Drake:
This PR adds the ability to configure the Max Distance parameter for Chrony and NTP. This will allow, among others, to sync Linux clients to a Windows NTP server.

PR #90, opened 2020-10-19
add jmespath as dependency for molecule

richm:
This should fix this problem:
```
   TASK [linux-system-roles.timesync : Set variable `timesync_services` with filtered uniq service names] ***

    fatal: [centos-6]: FAILED! => {"msg": "You need to install \"jmespath\" prior to running json_query filter"}

    fatal: [centos-7]: FAILED! => {"msg": "You need to install \"jmespath\" prior to running json_query filter"}
```
reported in https://github.com/linux-system-roles/timesync/pull/89#issuecomment-712104529

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -114,6 +114,7 @@ deps =
     molecule<3
     selinux
     wcwidth==0.1.9;python_version=="3.6"
+    jmespath
     -rmolecule_extra_requirements.txt
 runsyspycmd = {toxinidir}/.travis/runsyspycmd.sh
 


richm:
[citest bad]

richm:
@pcahyna looks like the fedora-32 failures are due to https://github.com/ansible/ansible/issues/71528 which is actually a systemd problem and should be fixed if we use an up-to-date fedora-32 image - that is - the current fix to only update the kernel does not solve this issue - we also need to update systemd

PR #104, opened 2021-02-23
use public: true to expose role parameters; ansible 2.8

richm:
Use `include_role` with `public: true` to expose role parameters
for testing.
Update `min_ansible_version` to 2.8.


spetrosi (approved):
LGTM

Issue #128, opened 2021-11-12
timesync failes on fedora35 trying to disable ntpd service because it is not installed


RobVerduijn:
Hello,

When applying the timesync role on fedora35 it tries to disable ntpd and it failed because it is not installed.

somehow it failed to check that ntpd was not installed.

ansible and collections versions
```
ansible [core 2.12.0]
  config file = /home/rob/code/playbooks/ansible.cfg
  configured module search path = ['/home/rob/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/rob/code/playbooks/venv/lib64/python3.10/site-packages/ansible
  ansible collection location = /home/rob/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/rob/code/playbooks/venv/bin/ansible
  python version = 3.10.0 (default, Oct  4 2021, 00:00:00) [GCC 11.2.1 20210728 (Red Hat 11.2.1-1)]
  jinja version = 3.0.2
  libyaml = True

ansible-galaxy collection list                                                                             127   master 

# /home/rob/.ansible/collections/ansible_collections
Collection                Version
------------------------- -------
RobVerduijn.lab           0.1.2  
ansible.netcommon         2.4.0  
ansible.posix             1.3.0  
ansible.utils             2.4.2  
ansible.windows           1.8.0  
community.crypto          2.0.0  
community.general         4.0.1  
community.libvirt         1.0.2  
community.windows         1.8.0  
fedora.linux_system_roles 1.10.1 
freeipa.ansible_freeipa   0.4.2  
theforeman.foreman        3.0.0  
theforeman.operations     0.3.0

```

spetrosi:
Hello, please share the playbook that you used so that we can reproduce the issue

RobVerduijn:
reproducer:
```
---
- name: reproducer
  hosts: poc.linux.lab
  gather_facts: true
  remote_user: root

  tasks:
    - name: include role
      include_role:
        name: fedora.linux_system_roles.timesync
```
see the attachement for the logs
[output.txt](https://github.com/linux-system-roles/timesync/files/7527311/output.txt)

spetrosi:
I am trying to reproduce this and in my tests Ansible skips the task `Disable ntpd`. What variables do you pass with the role?

RobVerduijn:
Hello,

I've run the play again now with debug statement at the beginning and  end 

see the attachment for results

```
- name: reproducer
[...]

PR #129, opened 2021-11-17
reject services which have a status == not-found

richm:


richm:
candidate fix for https://github.com/linux-system-roles/timesync/issues/128

richm:
The ansible-2.12 failures are fixed by https://github.com/linux-system-roles/timesync/pull/130 - the other failures are known

nhosoi (approved):
lgtm

tasks/main.yml
```diff
@@ -20,7 +20,9 @@
 - name: Set variable `timesync_services` with filtered uniq service names
   set_fact:
     timesync_services: "{{ ansible_facts.services.values() |
-          selectattr('name', 'defined') | map(attribute='name') |
+          selectattr('name', 'defined') |
+          rejectattr('status', 'match', '^not-found$') |
```
spetrosi, line 24: Looking at the fix in the service_fact module upstream, https://github.com/ansible/ansible/blob/stable-2.12/lib/ansible/modules/service_facts.py#L212, I think the following must be added also:
*  `rejectattr('status', 'match', '^masked$')`
* `rejectattr('status', 'match', '^failed$')`
Same workaround is required here:
https://github.com/linux-system-roles/storage/blob/master/tasks/main-blivet.yml#L54
I've opened AAP-1128 on Red Hat's Jira to cherry-pick this to ealrier Ansibles.

PR #134, opened 2022-01-07
Fix an issue if a service is listed by service_facts that does not have the 'status' property defined

Thulium-Drake:
When running this role I on a Debian ran into an issue where the ```service_facts``` module reports a bunch of things which have no ```status``` and breaks the subsequent task.

This PR will filter out such ```services```.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -21,7 +21,7 @@
   set_fact:
     timesync_services: "{{ ansible_facts.services.values() |
           selectattr('name', 'defined') |
-          selectattr('status', 'defined') |
+          rejectattr('status', 'undefined') |
           rejectattr('status', 'match', '^not-found$') |
           rejectattr('status', 'match', '^masked$') |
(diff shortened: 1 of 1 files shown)


Thulium-Drake:
After fixing this, I realized that there's more work to be done to make it Debian compatible, however, this fix is still useful on it's own :)

tasks/main.yml
```diff
@@ -21,6 +21,7 @@
   set_fact:
     timesync_services: "{{ ansible_facts.services.values() |
           selectattr('name', 'defined') |
+          rejectattr('status', 'undefined') |
```
richm, line 24: I don't think `undefined` is a valid test - I think you should use this instead
`selectattr('status', 'defined') |`

Thulium-Drake: Fair enough, I had that first time around (both worked). This is cleaner :)

richm:
[citest commit:6dbfe24e10868d1d1e4a81ff4e7eabec32efc23b]

PR #141, opened 2022-02-25
handle errors with stopping services

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=2058645
Not sure if it is a problem with rhel9 systemctl list-units, or a problem
with ansible-core 2.12 parsing that output, but service_facts reports
that ptp4l is a valid service, and the service module reports an error
trying to use state: stopped.  For example, here is the corresponding
systemctl status ptp4l when service_facts reports a valid service:
`Failed to get properties: Access denied`
Here is the systemctl list-units output
```
  ptp4l.service                          loaded    inactive dead    Precision Time Protocol (PTP) service
```
Here is the service_fact:
```
    "ptp4l.service": {
        "name": "ptp4l.service",
        "source": "systemd",
        "state": "stopped",
        "status": "inactive"
    },
```

The fix is to check the result of the `state: stopped` task.  If it is
a failure, but the MSG is `Could not find the requested service NAME`,
then ignore the failure.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -226,6 +226,11 @@
   when:
     - timesync_mode != 1 or timesync_ntp_provider != 'chrony'
     - "'chronyd' in timesync_services"
+  register: __disable_result
+  failed_when:
+    - __disable_result is failed
+    - not __disable_result.msg is match(
+      'Could not find the requested service chronyd:')
 
 - name: Disable ntpd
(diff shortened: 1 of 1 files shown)


richm:
[citest bad] - network issue with repo

spetrosi (approved):
lgtm

PR #143, opened 2022-04-13
support gather_facts: false; support setup-snapshot.yml (#52)

nhosoi:
Some users use `gather_facts: false` in their playbooks.  This changes
the role to work in that case, by gathering only the facts it requires
to run.
CI testing can be sped up by creating a snapshot image pre-installed
with packages.  tests/setup-snapshot.yml can be used by a CI system
to do this.

### DIFF
--- a/tests/setup-snapshot.yml
+++ b/tests/setup-snapshot.yml
@@ -1,0 +1,26 @@
+- hosts: all
+  tasks:
+    - name: Set platform/version specific variables
+      include_role:
+        name: linux-system-roles.timesync
+        tasks_from: set_vars.yml
+        public: true
+
+    - name: Install test packages to put them in the local cache
+      package:
+        name:
+          - chrony
+          - ntp
+          - linuxptp
+        state: present
+      no_log: true
+      ignore_errors: true
+
+    - name: Uninstall test packages
+      package:
+        name:
+          - chrony
+          - ntp
+          - linuxptp
+        state: absent
+      no_log: true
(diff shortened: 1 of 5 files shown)


tests/setup-snapshot.yml
```diff
@@ -0,0 +1,12 @@
...
+  tasks:
+    - name: Set platform/version specific variables
+      include_role:
+        name: linux-system-roles.timesync
+        tasks_from: set_vars.yml
+        public: true
+
+    - name: Install test packages
+      package:
+        name: "{{ timesync_ntp_provider_os_default }}"
```
richm, line 11: This works, but only by accident, that the name of the provider is the same as the name of the OS package.
Unfortunately, it doesn't look like the timesync role is really suitable for using setup-snapshot - there is no clearly defined list of packages, and there is a *lot* of logic involved to determine the correct packages to install.
I guess what you could do in setup-snapshot is to install `chrony`, and `ntp`, and `linuxptp`, then remove them.  Since not all of these packages will be available on all platforms, just set `no_log: true` and `ignore_errors: true` when installing them (you should not need to ignore errors upon removal).  At least that way the packages will be in the local cache so at least we will get a slight performance improvement.

vars/main.yml
```diff
@@ -4,3 +4,12 @@ __timesync_chrony_version: "{{
...
   if 'ntp' in ansible_facts.packages else '0' }}"
+
+# ansible_facts required by the role
+__timesync_required_facts:
+  - distribution
+  - distribution_major_version
+  - distribution_version
+  - os_family
+  - packages
+  - services
```
richm, line 15: I don't think packages and services should be in this list - they are not required to determine which packages to install - there are separate `package_facts:` and `service_facts:` modules for those.

nhosoi:
Thank you, @richm. Updated the pr. If it looks ok, I'm making the two commits into one.

richm (approved):
lgtm

nhosoi:
> lgtm

Thank you, @richm. Two commits are merged into one, now.

PR #145, opened 2022-06-16
WIP - add before, before suite, after, and after suite playbooks

richm:
This adds playbooks to run at certain points in tests:

tests/_before_test_suite.yml is intended to be run before the
first test (and before tests/_before_test.yml).  The intention
is to save any state (config files, services, packages) that need
to be restored after each test, or at the end of the test suite.

tests/_after_test.yml is intended to be run after each test - it
restores the state of the system to allow the next test to run

tests/_after_test_suite.yml - clean up anything left over - so
the next test suite can run cleanly

tests/_before_test.yml - not provided, but would be intended to
run before each test - not really useful for timesync

tests/vars/main.yml - vars shared by test playbooks

This will require support in CI to take advantage of this.  For
example, when creating a batch file like this:
```
... -- tests/_before_test_suite.yml tests/_before_test.yml tests/tests_1.yml tests/_after_test.yml
... -- tests/_before_test.yml tests/tests_2.yml tests/_after_test.yml
...
... -- tests/_before_test.yml tests/tests_N-1.yml tests/_after_test.yml
... -- tests/_before_test.yml tests/tests_N.yml tests/_after_test.yml tests/_after_test_suite.yml
```

### DIFF
--- a/tests/_after_test.yml
+++ b/tests/_after_test.yml
@@ -20,11 +20,15 @@
     - name: Restore services to correct state
       service:
         name: "{{ __item['name'] }}"
-        state: "{{ (__item['state'] == 'running') | ternary('restarted', 'stopped') }}"
+        state: "{{ (__item['state'] == 'running') |
+          ternary('restarted', 'stopped') }}"
         enabled: "{{ __item['status'] == 'enabled' }}"
       loop: "{{ __timesync_services }}"
       ignore_errors: true
       vars:
-        __system_services: "{{ __timesync_services_raw['content'] | b64decode | from_json }}"
-        __sys_item: "{{ __system_services | selectattr('name', 'match', '^' ~ item ~ '$') }}"
-        __item: "{{ __sys_item[0] if __sys_item else {'name': item, 'state': 'stopped', 'status': 'disabled'} }}"
+        __system_services: "{{ __timesync_services_raw['content'] |
+          b64decode | from_json }}"
+        __sys_item: "{{ __system_services |
+          selectattr('name', 'match', '^' ~ item ~ '$') }}"
+        __item: "{{ __sys_item[0] if __sys_item else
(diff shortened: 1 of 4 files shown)


richm:
This is just a proposal for now - do not merge

nhosoi:
I have a couple of questions...
1) So far, the timesync test failure is observed only on rhel-6. Do you think we will run the _before and _after_test.yml only when the managed host is rhel-6?
2) I also think making the save and restore tasks a separate playbook is cleaner compared to making them a task yaml file to be imported (as being done in the metrics role); the former is slower than the latter. Maybe, it's not an issue since the duration of the timesync tests is rather short, but do we care the overhead?

richm:
> I have a couple of questions...
> 
>     1. So far, the timesync test failure is observed only on rhel-6. Do you think we will run the _before and _after_test.yml only when the managed host is rhel-6?

I think it should be done for every role, for every platform, to catch cases like this.  However, so far, we have only ever had a problem with the timesync role, so it isn't a high priority for now.

That is - every role that modifies config files and services (and packages), that doesn't have a `cleanup` block in the test, could potentially have this problem.  Either all of the affected roles have clean-up blocks, or we just haven't run a test that easily gets into this situation.

> 
>     2. I also think making the save and restore tasks a separate playbook is cleaner compared to making them a task yaml file to be imported (as being done in the metrics role); the former is slower than the latter. Maybe, it's not an issue since the duration of the timesync tests is rather short, but do we care the overhead?

I don't know - I guess we would have to measure it.  The big advantage of having separate playbooks is that we could run ansible-playbook with the `_after` playbooks even if the test fails, to ensure that we always clean up after the test.  If you have the before and after tasks in an imported tasks file, you have to be careful to have `block/always` or you might miss some cleanup.

PR #154, opened 2022-08-03
Update chrony.conf.j2 with type casting

mto79:
Variables are defined as string in Jinja2 template and the operators 
- on line 26 gives '<'  not supported between instances of 'str' and 'float
- on line 43 gives '>' not supported between instances of 'str' and 'float

changed template with type casting.
Had the problem on target RHEL 8.6  en RHEL9.0. In RHEL9.0 setting jinja2_native did work.

Tests done with:
ansible node has ansible 2.9.27/pyhton 3.8.10)
ansible node has ansible 2.9.27/pyhton 3.8.10)

Never had troubles with this system role and very unclear how the vars became strings in Jinja2 templating ..

### DIFF
--- a/templates/chrony.conf.j2
+++ b/templates/chrony.conf.j2
@@ -21,9 +21,9 @@ value['hostname'] }}{{
 sourcedir {{ timesync_chrony_dhcp_sourcedir }}
 
 {% endif %}
-{% if timesync_step_threshold != 0.0 %}
+{% if timesync_step_threshold|float != 0.0 %}
 # Allow the system clock to be stepped in the first three updates.
-makestep {{ timesync_step_threshold if timesync_step_threshold > 0.0 else '1.0' }} 3
+makestep {{ timesync_step_threshold if timesync_step_threshold|float > 0.0 else '1.0' }} 3
 
 {% endif %}
 # Enable kernel synchronization of the real-time clock (RTC).
@@ -40,7 +40,7 @@ hwtimestamp {{ interface }}{% if __timesync_chrony_version is version('3.1', '>=
 {% endfor %}
 
 {% endif %}
-{% if timesync_min_sources > 1 %}
+{% if timesync_min_sources|int > 1 %}
 # Increase the minimum number of selectable sources required to adjust
 # the system clock.
 minsources {{ timesync_min_sources }}
--- a/templates/chrony.conf.j2
+++ b/templates/chrony.conf.j2
@@ -21,9 +21,9 @@ value['hostname'] }}{{
 sourcedir {{ timesync_chrony_dhcp_sourcedir }}
 
 {% endif %}
-{% if timesync_step_threshold|float != 0.0 %}
+{% if timesync_step_threshold | float != 0.0 %}
 # Allow the system clock to be stepped in the first three updates.
-makestep {{ timesync_step_threshold if timesync_step_threshold|float > 0.0 else '1.0' }} 3
+makestep {{ timesync_step_threshold if timesync_step_threshold | float > 0.0 else '1.0' }} 3
 
 {% endif %}
 # Enable kernel synchronization of the real-time clock (RTC).
@@ -40,7 +40,7 @@ hwtimestamp {{ interface }}{% if __timesync_chrony_version is version('3.1', '>=
 {% endfor %}
 
 {% endif %}
-{% if timesync_min_sources|int > 1 %}
+{% if timesync_min_sources | int > 1 %}
 # Increase the minimum number of selectable sources required to adjust
 # the system clock.
 minsources {{ timesync_min_sources }}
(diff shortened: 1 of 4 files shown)


richm:
> Variables are defined as string in Jinja2 template and the operators
> 
>     * on line 26 gives '<'  not supported between instances of 'str' and 'float
> 
>     * on line 43 gives '>' not supported between instances of 'str' and 'float
> 
> 
> changed template with type casting. Had the problem on target RHEL 8.6 en RHEL9.0. In RHEL9.0 setting jinja2_native did work.
> 
> Tests done with: ansible node has ansible 2.9.27/pyhton 3.8.10) ansible node has ansible 2.9.27/pyhton 3.8.10)
> 
> Never had troubles with this system role and very unclear how the vars became strings in Jinja2 templating ..

Sounds similar to https://github.com/willshersystems/ansible-sshd/issues/188 which turned out to be a bug in jinja 3.0, which was fixed in jinja 3.1.

Did you notice this problem happening after you upgraded to fix the issue in https://github.com/linux-system-roles/kernel_settings/issues/95#issuecomment-1204534125 ?

Note that in rhel 8.7 and rhel 9.1, ansible-core will be upgraded from 2.12 to 2.13.  Along with this change, ansible-core will bundle its own jinja2 3.1.x instead of relying on the system python jinja2 library that is used in rhel 8.6/9.0.

mto79:
The problem was before upgrading for the other issue with the kernel_settingss. But also with an another ansible control node and rhel 8.6 test environment. In one case with Jinja version jinja version = 2.11.3 and the other with python3-jinja2-3.0.1-2.fc35.noarch.

Nice to know =>
ansible-core will bundle its own jinja2 3.1.x instead of relying on the system python jinja2 library that is used in rhel 8.6/9.0.

richm:
> The problem was before upgrading for the other issue with the kernel_settingss. But also with an another ansible control node and rhel 8.6 test environment. In one case with Jinja version = 2.11.3 and the other with python3-jinja2-3.0.1-2.fc35.noarch.
> 
> Nice to know => ansible-core will bundle its own jinja2 3.1.x instead of relying on the system python jinja2 library that is used in rhel 8.6/9.0.

hmm - I'm not really sure how to proceed here - I suspect other system roles will have the same problem - on the other hand, not sure if other system roles rely on `int` and `float` comparisons in jinja templates.  Also, it doesn't seem right to have to change the system role when this seems very clearly a jinja problem which is fixed in jinja 3.1

mto79:
> > The problem was before upgrading for the other issue with the kernel_settingss. But also with an another ansible control node and rhel 8.6 test environment. In one case with Jinja version = 2.11.3 and the other with python3-jinja2-3.0.1-2.fc35.noarch.
> > Nice to know => ansible-core will bundle its own jinja2 3.1.x instead of relying on the system python jinja2 library that is used in rhel 8.6/9.0.
> 
> hmm - I'm not really sure how to proceed here - I suspect other system roles will have the same problem - on the other hand, not sure if other system roles rely on `int` and `float` comparisons in jinja templates. Also, it doesn't seem right to have to change the system role when this seems very clearly a jinja problem which is fixed in jinja 3.1

Totally agree (just have read https://github.com/willshersystems/ansible-sshd/issues/188) They just made comparisons stricter, i saw that also in the newer Python versions (Python2.7 is less strict) .. also saw it the versions( '2', '<')  that had to be compared with ansible_facts.distribution_major_version instead of ansible_distribution_major_version.

mto79:
Maybe a note in the requirements for ansible version and or jinja version ?

A requirements.txt like
# Tests run on the previous and current (latest) version of Ansible.
ansible>=2.10
# Some Jinja2 filters are used that are available in the newer releases.
jinja2>=2.11.2

richm:
[citest]

richm:
I have updated our Ansible scan script to look for public api variables used in numeric operations - https://github.com/linux-system-roles/auto-maintenance/pull/181 - this is what it shows with the timesync role:
```
[timesync_step_threshold] [gteq] [0.0] at timesync templates/phc2sys.sysconfig.j2:3
[timesync_step_threshold] [gteq] [0.0] at timesync templates/ntp.conf.j2:32
[timesync_min_sources] [gt] [1] at timesync templates/ntp.conf.j2:37
[timesync_max_distance] [ne] [0] at timesync templates/ntp.conf.j2:43
[timesync_step_threshold] [gteq] [0.0] at timesync templates/ptp4l.conf.j2:16
[timesync_step_threshold] [ne] [0.0] at timesync templates/chrony.conf.j2:24
[timesync_step_threshold] [gt] [0.0] at timesync templates/chrony.conf.j2:26
[timesync_min_sources] [gt] [1] at timesync templates/chrony.conf.j2:43
[timesync_max_distance] [ne] [0] at timesync templates/chrony.conf.j2:77
```
Please fix all of these with your PR

mto79:
Nice! Just update the PR. Only doubt on timesync_max_distance to be int or float .. https://chrony.tuxfamily.org/doc/3.4/chrony.conf.html#maxdistance. Saw values of 16.0 etc .. so choosen for float.

mto79:
nope must be integer, following the comparision and further reading

richm:
[citest]

richm:
thanks for the update - lgtm - will make sure tests pass

richm (approved):

PR #170, opened 2023-01-18
Add check for non-inclusive language; 

nhosoi:
Add check for non-inclusive language

Add a check for usage of terms and language that is considered non-inclusive. We are using the woke tool for this with a wordlist that can be found at
https://github.com/linux-system-roles/tox-lsr/blob/main/src/tox_lsr/config_files/woke.yml

Clean up / Workaround non-inclusive words
- CHANGELOG.md
- README.md
- handlers/main.yml
- tasks/main.yml
- templates/ptp4l.conf.j2
- templates/timemaster.conf.j2
- tests/tests_ptp_multi.yml
- vars/CentOS_6.yml
- vars/CentOS_9.yml
- vars/Debian.yml
- vars/RedHat_6.yml
- vars/RedHat_9.yml
- vars/default.yml

Note: woke action depends on https://github.com/linux-system-roles/tox-lsr/pull/103 for skipping the check on the filename timemaster.conf.j2.

### DIFF
--- a/.github/workflows/woke.yml
+++ b/.github/workflows/woke.yml
@@ -1,0 +1,19 @@
+---
+# yamllint disable rule:line-length
+name: Check for non-inclusive language
+on:  # yamllint disable-line rule:truthy
+  - pull_request
+jobs:
+  woke:
+    name: woke
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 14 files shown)


richm:
[citest]

PR #178, opened 2023-01-30
Fix jinja formatting issue

richm:
Fix for the following warning:
```
TASK [linux-system-roles.timesync : Disable timemaster] ************************
[WARNING]: conditional statements should not include jinja2 templating
delimiters such as {{ }} or {% %}. Found: '{{ __timemstr }}' in
timesync_services
```
`when`, `that`, and similar are already interpreted as Jinja expressions
and do not require the use of {{ }}

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -304,12 +304,12 @@
     enabled: false
   when:
     - timesync_mode != 3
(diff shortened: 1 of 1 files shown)


richm:
[citest]

PR #135, opened 2022-01-07
Initial version for Debian

Thulium-Drake:
With these edits (and the one from #134 ) I was able to succesfully configure Chronyd on my Debian system.

Feedback is appreciated! :-)

PR #187, opened 2023-03-29
Update chrony.conf location for Debian

Thulium-Drake:
The config file for chrony is in a different location on Debian (I missed that in #135 :-) )


richm:
[citest]

PR #216, opened 2023-10-09
refactor: Add CentOS 7-8, RedHat 7-8 and AlmaLinux 8-9 var files

scaronni:
Enhancement: Add variables for additional operating systems.

Reason: They are missing, so values in `default.yml` are used.

Result: Everything works as expected in the added operating systems.

The default values are fine for all the operating systems listed, so in my opinion the best approach would be to get rid of all of them, leaving only `default.yml` and simplify. But in the meanwhile, this makes these additional OSes "compliant".


richm:
[citest]

richm (approved):
the fedora 38 failures are unrelated - @mlichvar any ideas?

Issue #48, opened 2019-09-20
Adding the timesync_ntp_allow variable for setuping a server as source server


florian0410:
We started to use the role in a private cloud and we are using a single server that connect to externals ntp servers for synchronization while other are offline.

We would like to use it as an NTP server and client at the same time.
From what I read we need to use the configuration **allow** [https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system_administrators_guide/sect-understanding_chrony_and-its_configuration](url)

We should add a new variable to handle this in **timesync/templates/chrony.conf.j2**, for example:

```yaml
timesync_allow_network:
  - ip_net: 127.0.1.0/24
```

Which would result in the vars like this:

```yaml
{% if timesync_allow_network is defined %}
{% for value in timesync_allow_network %}
{{ 'allow' if 'ip_net' in value and value['ip_net'] }} {{value['ip_net'] }}
{% endfor %}
{% endif %}
```

Resulting in in the configuration as follow :

```
allow 127.0.1.0/24
```

mlichvar:
As currently documented, this role is meant only for configuring NTP clients and/or PTP slaves. It might make sense to extend it into a NTP server, but we need to consider all supported use cases and how it will work with future improvements. For instance, if the role is configured with a single PTP domain, there is no ntpd/chronyd currently running. Should it be started? Or would it be better to rely on a separate "NTP server" role?

PR #46 is a similar issue.

richm:
Use `timesync_chrony_custom_settings`

PR #217, opened 2023-10-09
Add allow directive in configuration file

scaronni:
Enhancement: Add `allow` directive in `chrony.conf` configuration file and relevant documentation.

Reason: This directive is required to configure `chrony` in server mode, otherwise it operates purely as an NTP client.

Result: `chrony` starts in server mode and listens to port 123 as described in the `chrony.conf(5)` `man` page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,14 @@ timesync_ntp_hwts_interfaces: ["*"]
 # package specific to the system and its version will be selected.
 timesync_ntp_provider: chrony
 
+# List of allowed network, ips and ranges that can request time.
+# This is required to operate chrony as a server, without it chronyd operates
+# purely as an NTP client.
+timesync_allow:
+  - 1.2.3.4
+  - 3.4.5.0/24
+  - all
+
 # Sometimes administrators might need extended configurations for chrony which
 # are not covered by the predefined settings provided by this role.
 # 'timesync_chrony_custom_settings' allows to define a list of custom settings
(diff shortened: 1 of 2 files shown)


mlichvar:
As described in the README, this role is intended for configuration of NTP and PTP clients.

You can add `allow` to `timesync_chrony_custom_settings` if you really need to, but I'm not sure if want to complicate the role with setting like that.

mlichvar:
Have you tried adding `allow` to `timesync_chrony_custom_settings` ?

There are many roles for configuring chrony. This one is specifically for clients. It supports also ntpd, which has different configuration (it cannot operate purely as a client) so this would be confusing. There are many other server-specific directives, e.g. for NTS, which is already support as a client, so this would be incomplete.

richm:
> Have you tried adding `allow` to `timesync_chrony_custom_settings` ?
> 
> There are many roles for configuring chrony. This one is specifically for clients. It supports also ntpd, which has different configuration (it cannot operate purely as a client) so this would be confusing. There are many other server-specific directives, e.g. for NTS, which is already support as a client, so this would be incomplete.

@scaronni did you try this, and did it work for you?  If so, can we close this PR?

scaronni:
Yes it works, but not really evident. But i've seen this has been asked in the past with the same answer (https://github.com/linux-system-roles/timesync/issues/48), so let's close it.

PR #246, opened 2024-07-01
fix: Don't use chrony-dhcp sourcedir on EL8 systems

mlichvar:
On EL8 systems the chrony package uses a helper script which adds servers from DHCP with the "chronyc add" command. The /run/chrony-dhcp directive is used on EL9 and later.


richm (approved):
ok - looks like the change was already made for AlmaLinux_8.yml and CentOS_8.yml

PR #275, opened 2025-03-20
ci: Create inventory for wrapper test with inventory_hostname

spetrosi:
Enhancement: Create inventory for wrapper test with inventory_hostname

Reason: Multihost tests have an inventory defining many nodes, then run playbooks with `--limit <node>`. Hence, the role must build inventory for wrapper tests only with the node that it runs against.

Result: Inventory that is created for wrapper tests contains a single node like in https://github.com/linux-system-roles/kdump/blob/main/tests/inventory.yaml.j2

### DIFF
--- a/tests/inventory.yaml.j2
+++ b/tests/inventory.yaml.j2
@@ -1,1 +1,11 @@
-{{ { 'all': { 'hosts': hostvars } } | to_yaml }}
+all:
+  hosts:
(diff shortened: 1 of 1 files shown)


spetrosi:
[citest]

richm (approved):

spetrosi:
[citest]

PR #277, opened 2025-03-31
feat: add support for timesync_ntp_ip_family

richm:
Feature: Add support for timesync_ntp_ip_family to allow setting the `-4`
or `-6` OPTIONS in the chronyd or ntpd sysconfig file.

Reason: When IPv6 is disabled on the node, you must tell chronyd to
only listen for IPv4 using OPTIONS="-4" in the sysconfig file.
Otherwise, chronyd will log error messages when binding to IPv6 sockets
It's also useful to prevent the client from using IPv6 servers when IPv4
is known to work better (e.g. IPv6 over a tunnel).

Result: chronyd and ntpd can be configured to work correctly, and
the services will not log errors, when IPv6 (or IPv4) is disabled
on the node.

### DIFF
--- a/templates/chronyd.sysconfig.j2
+++ b/templates/chronyd.sysconfig.j2
@@ -1,4 +1,6 @@
 {{ ansible_managed | comment }}
 {{ "system_role:timesync" | comment(prefix="", postfix="") }}
 
-OPTIONS=""
+OPTIONS="{{ ' -4' if timesync_ntp_ip_family == 'IPv4'
+    else ' -6' if timesync_ntp_ip_family == 'IPv6'
+    else '' }}"
--- a/README.md
+++ b/README.md
@@ -121,6 +121,18 @@ timesync_chrony_custom_settings:
 # the role will fail to ensure the reboot requirement is not overlooked.
 # For non-transactional update systems, this variable is ignored.
 timesync_transactional_update_reboot_ok: true
+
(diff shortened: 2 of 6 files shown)


richm:
[citest]

templates/chronyd.sysconfig.j2
```diff
@@ -1,4 +1,6 @@
 {{ ansible_managed | comment }}
 {{ "system_role:timesync" | comment(prefix="", postfix="") }}
 
-OPTIONS=""
+OPTIONS="{{ ' -4' if timesync_ntp_ip_family == 'IPv4'
+    else ' -6' if timesync_ntp_ip_family == 'IPv6'
+    else '' }}"
```
spetrosi, line 6: Previously it was "" instead of empty string. Tests pass but I do not know if this is covered.

mlichvar:
The code looks good to me. I'd just suggest to improve the commit message. It's not that chronyd doesn't work on a IPv6-disabled host without -4, but it logs error messages when binding to the IPv6 sockets fails. Adding the -4 option disables those sockets and there are no error messages. It's also useful to prevent the client from using IPv6 servers when IPv4 is known to work better (e.g. IPv6 over a tunnel).

mlichvar (approved):

richm:
[citest]

richm:
[citest]

PR #286, opened 2025-04-23
test: ensure tests cleanup and restore the system state

richm:
Ensure the tests cleanup and restore the system state.
Refactor tests to use a `block/always` so that cleanup can
occur in the `always` block.
Add a tests/tasks/cleanup.yml tasks file which is run from
the `always` block of the test, and include this task file
to be run from the `always` block of tests that require cleanup.


richm:
[citest]

richm:
[citest]

PR #288, opened 2025-04-24
test: add test for secccomp on el9/el10

richm:
Add a test to ensure that running the role with no
variables does not remove the default OPTIONS for seccomp.

This also required adding a vars file for Fedora


richm:
This is the test for https://github.com/linux-system-roles/timesync/pull/279

richm:
[citest]

PR #345, opened 2026-04-27
feat: add role fingerprints to syslog

richm:
Feature: Add a fingerprint string to the system log to indicate when the role began
successfully, and when the role finished successfully.  The fingerprint string indicates
the role name, a timestamp, and the platform.

Reason: Users can see when the role was used and if it was used successfully.  This
information from the system log can be collected by log scanners and aggregators
for further analysis.

Result: The role logs fingerprints to the system log.

This also adds a test to check if the fingerprints were written upon a successful
role invocation.

## Summary by Sourcery

Add syslog fingerprinting for the timesync role and verify fingerprints via journal inspection in tests.

New Features:
- Introduce sr_fingerprint Ansible module to log fingerprint messages to syslog with timestamps.
- Record begin and success fingerprint messages for the linux-system-roles.timesync role including Ansible and platform metadata.

Tests:
- Add integration test that verifies role fingerprint messages are written to the system journal when syslog is available.

### DIFF
--- a/.sanity-ansible-ignore-2.10.txt
+++ b/.sanity-ansible-ignore-2.10.txt
@@ -1,2 +1,3 @@
 plugins/modules/timesync_provider.sh shebang
 plugins/modules/timesync_provider.yml validate-modules:missing-gplv3-license
+plugins/modules/sr_fingerprint.py validate-modules:missing-gplv3-license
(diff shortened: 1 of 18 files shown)


richm:
[citest]

richm:
[citest]

Issue #321, opened 2026-01-26
Feature Request: Support for `key` directive in chrony.conf


GianniBYoung:
I need to be able to configure servers with the `key` field

```
# /etc/chrony.conf
server <hostname> key <keyid> maxpoll 10
```` 

In order to do this currently I need to pass the server configure via custom settings:

```yaml
timesync_chrony_custom_settings:
  - "server <hostname> key <keyid> maxpoll 10"
```

And I also need to push the key files to `keyfile /etc/chrony.keys` via a custom task.

```
# /etc/chrony.keys
<keyid> <hashing algorithm> <password hash>
25 SHA1 HEX:1dc764e0791b11fa67efc7ecbc4b0d73f68a070c
```

I would like to be able to handle this scenario natively with the role configuration

richm:
@GianniBYoung We do not have plans to support this authentication method - according to https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/7/html-single/system_administrators_guide/index#s1-Authentication_Options_for_NTP:

> NTPv4 added support for the Autokey Security Architecture, which is based on public asymmetric cryptography while retaining support for symmetric key cryptography. The Autokey protocol is described in RFC 5906 Network Time Protocol Version 4: Autokey Specification. Unfortunately, it was found later that the protocol has serious security issues, and thus Red Hat strongly recommends to use symmetric keys instead. The man page ntp_auth(5) describes the authentication options and commands for ntpd

The recommended method is Network Time Security (NTS) https://chrony-project.org/faq.html "2.6. How can I make the system clock more secure?". 

Are you able to use NTS?  It has been supported since chrony version 4.0, and this is available in EL8 and later.

richm:
@GianniBYoung please review and try https://github.com/linux-system-roles/timesync/pull/357

PR #357, opened 2026-07-14
feat: Support key and keyfile for authentication

richm:
Feature: Allow users to specify an authentication key for a server using the
new `key` parameter for items in the `timesync_ntp_servers` list.  The `key`
value is as described at https://chrony-project.org/doc/latest/chrony.conf.html#keyfile
and includes the key ID, the optional type, and the key string.  Users are strongly
encouraged to use Ansible Vault to encrypt the key.

Reason: Users need to be able to specify authentication keys for secure
time synchronization.

Result: Users can configure secure time synchronization.

Assisted-by: Cursor using models Cursor Grok 4.5, Composer 2.5, Sonnet 5.

## Summary by CodeRabbit

* **New Features**
  * Added Chrony authentication support with per-server `key` values, automatically applied to generated `pool`/`server` entries.
  * Added configurable secure logging and an option to remove the Chrony keyfile when no keys are configured.
  * Added support for customizing the Chrony keyfile path.
* **Documentation**
  * Updated role documentation for the `key` format and new configuration options, including related error cases.
* **Bug Fixes**
  * Improved Chrony configuration rendering for keyfile placement and key handling.
* **Tests**
  * Expanded Chrony key generation/configuration test coverage, including failure scenarios and keyfile retention/removal behavior.


richm:
[citest]

richm:
[citest]

richm:
[citest]

tests/tests_keys.yml
```diff
@@ -0,0 +1,165 @@
...
+              - __keys_chrony_conf is search('172\.16\.124\.5.* key 30')
+              - __keys_chrony_conf is search('172\.16\.124\.6.* key 30')
+          when: ansible_version.full is version('2.10', '>=')
+
+        # the unique filter does not preserve order on ansible 2.9
+        - name: Check chrony.conf key settings on ansible 2.9
+          assert:
+            that:
+              - __keys_chrony_conf is search('keyfile ' ~ timesync_chrony_keyfile_path)
+              - __keys_chrony_conf is search('172\.16\.124\.1.* key [1-3]0')
```
spetrosi: Why does ansible 2.9 support this? It used to support some more direct python modules invocations?

spetrosi (approved):

tests/tests_keys.yml
richm: not sure what you mean.  This is needed because the `unique` filter in ansible 2.9 used at https://github.com/linux-system-roles/timesync/pull/357/changes#diff-3d0ff1709ca48add100327bb2a468e6c508fb92a159c64c4f99ad1df89d9bddeR167 does not preserve the order - so the key IDs aren't 10, 20, 30 in the same order as they are listed in the test in `timesync_ntp_servers` - in later versions of ansible the `unique` filter preserves the order, so we can be guaranteed that the key IDs will be in the same order as the servers listed in `timesync_ntp_servers`

richm:
Fixes https://github.com/linux-system-roles/timesync/issues/321

mlichvar (commented):
The key IDs need to be specified by the user, so they match the server configuration. This seems quite a lot of code for such a minor feature.

richm:
> The key IDs need to be specified by the user, so they match the server configuration.

Ah, ok.  I thought the key ID was just used as an index into the chrony.keys "table".  That is, I thought the use case was this: As an administrator configuring the client side of secure timesync, I need to configure the client to use the same authentication key as used by the server, which is just the "[type ]keystring".  So I figured "hey, the user doesn't care about the key ID since it is just an index into the chrony.keys table - they just want to associate a key with a server, so the role can auto-generate the key IDs".

So the key ID is significant?  That will definitely change the implementation, and possibly simplify somewhat.

> This seems quite a lot of code for such a minor feature.

It may be "minor" but it is security related, which invokes several considerations. 

* The vars files changes are to add variables for the keyfile path and group.  The reason why so many files have to change for this is because the set_vars implementation uses the `first_found` method https://github.com/linux-system-roles/timesync/blob/main/tasks/set_vars.yml#L38 instead of the more flexible method used in newer roles https://github.com/linux-system-roles/podman/blob/main/tasks/set_vars.yml#L39
* vault is used because we really want to emphasize that customers should use vault to encrypt the key values
* `timesync_secure_logging` and `no_log` was added because the role must not leak key data into the logs
* the code to parse the keys and generate the key IDs is because of the reasons listed above
* the main.yml code that uses the keys was moved into a block so that we can ensure the variables holding the key material are cleared and not leaked into the global namespace/scope

If you can suggest a better way to do this, I would appreciate it

richm:
* added `timesync_remove_keyfile_if_no_keys` because otherwise there is no way to remove a `chrony.keys` file that is no longer used, and it seems unsafe to just always remove it if there are no keys specified in the `timesync_ntp_servers` list without user authorization

tests/tests_keys.yml
richm: The new implementation doesn't do this and works the same on 2.9 and later.

richm:
@mlichvar the new implementation takes a `key` as the entire string including the key ID

richm:
[citest]

richm:
[citest_bad]

richm:
[citest]

richm:
[citest_bad]

mlichvar:
> @mlichvar the new implementation takes a `key` as the entire string including the key ID

Looks good to me. Thanks.

What will happen if there are two servers using the same key ID but different key?

richm:
> > @mlichvar the new implementation takes a `key` as the entire string including the key ID
> 
> Looks good to me. Thanks.
> 
> What will happen if there are two servers using the same key ID but different key?

The latest commit addresses this:

* the role will issue an error if the user attempts to use multiple keys which have different key IDs but the same key value, or the same key ID for different key values
* there are tests for these cases
* the README explains this

richm:
[citest]

mlichvar:
The new "Ensure the same key value is not used with different IDs" check doesn't seem right to me. It might not make much sense to use different IDs for the same key, but I think it should not cause the role to fail.

richm:
> The new "Ensure the same key value is not used with different IDs" check doesn't seem right to me. It might not make much sense to use different IDs for the same key, but I think it should not cause the role to fail.

How about a warning in that case?

mlichvar:
Yes, that would be fine I think.

richm:
[citest]

richm:
> Yes, that would be fine I think.

ok - check the latest commit - ok to merge?

mlichvar:
"Each unique key value should have a unique ID." in the warning seems wrong. I think the warning should just hint it might be a misconfiguration, possibly also on the server side (two different servers using same key).

Also the new text in documentation should be fixed to not say it's an error.

richm:
> "Each unique key value should have a unique ID." in the warning seems wrong. I think the warning should just hint it might be a misconfiguration, possibly also on the server side (two different servers using same key).
> 
> Also the new text in documentation should be fixed to not say it's an error.

Done.

richm:
[citest]

richm:
[citest]

richm:
[citest]

richm:
[citest]

PR #366, opened 2026-08-07
test: ensure fingerprint test works in check mode

richm:
The fingerprint journal messages and the fingerprint log file
are not written in check mode.  Ensure that the correct messages
are written in check mode.

## Summary by CodeRabbit

* **Bug Fixes**
  * Verification and log collection now run only when applicable, preventing failures in check mode or when `/dev/log` is unavailable.
  * Improved check-mode validation to confirm expected time synchronization messages and check-mode status.

### DIFF
--- a/tests/tests_default.yml
+++ b/tests/tests_default.yml
@@ -17,25 +17,10 @@
       vars:
         __timesync_write_log_file: true
 
-    - name: Get fingerprint entries from journal
-      ansible.builtin.shell:
-        executable: /bin/bash
-        cmd: >-
-          set -eo pipefail;
-          journalctl --since "{{ __journal_start_time }}" --no-pager |
-          grep -v " Invoked with" |
-          grep "sr_fingerprint.*role_name=timesync"
-      register: __register_journal_fingerprints
-      changed_when: false
-      when: __register_dev_log.stat.exists
-
-    - name: Check that the log file was written
-      ansible.builtin.slurp:
-        path: /var/log/sysroles.jsonl
-      register: __register_log_file
-
     - name: Verify log file and journal fingerprints
-      when: __register_dev_log.stat.exists
+      when:
+        - not ansible_check_mode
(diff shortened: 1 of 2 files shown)


richm:
[citest]

richm:
@mlichvar getting these errors on fedora 43 and 44 with test tests_ntp_ptp.yml:

```
Aug 07 19:42:04 managed-node3 (timemaster)[27656]: timemaster.service: Failed to set up mount namespacing: /var/lib/chrony: No such file or directory
Aug 07 19:42:04 managed-node3 (timemaster)[27656]: timemaster.service: Failed at step NAMESPACE spawning /usr/sbin/timemaster: No such file or directory
```

any idea?

mlichvar:
@richm , it seems this test failure is caused by a change made in the chrony package, which delayed creation of the /var/lib/chrony directory from the package installation to the first start of the chronyd service. This breaks the timemaster service if chronyd wasn't started before. I'll see how this can be fixed without adding a dependency on the chrony package.

richm:
@mlichvar getting these errors on fedora 43 and 44 with test tests_ntp_ptp.yml:

```
Aug 07 19:42:04 managed-node3 (timemaster)[27656]: timemaster.service: Failed to set up mount namespacing: /var/lib/chrony: No such file or directory
Aug 07 19:42:04 managed-node3 (timemaster)[27656]: timemaster.service: Failed at step NAMESPACE spawning /usr/sbin/timemaster: No such file or directory
```

any idea?

> @richm , it seems this test failure is caused by a change made in the chrony package, which delayed creation of the /var/lib/chrony directory from the package installation to the first start of the chronyd service. This breaks the timemaster service if chronyd wasn't started before. I'll see how this can be fixed without adding a dependency on the chrony package.

We could add this to the role - ensure that `/var/lib/chrony` exists before starting timemaster.  I don't know if that would cause problems when chronyd starts up and finds the directory already exists, but it shouldn't.

